PR #6, opened 2022-03-25
Issue #5 Remove nightly toolchain since it should not be needed anymore

### LINKED ISSUE #5: Remove nightly toolchain since it should not be needed anymore
from @str4d 

> aarch64-apple-darwin  is a Tier 2 supported platform now, so it should have stable binaries.

pacu:
Closes #5 
Nightly toolchain builds are causing false positive compiler errors.
We will remove them from the Makefile since M1 is now fully supported
latest stable


dh-ecc (approved):
LGTM and builds fine on my M1. This change really helps reduce complexity of the build 😄.

Might be able to build on this for a couple more simplifications.

PR #7, opened 2022-03-30
Add min_confirmations argument to balance & spend methods.

### LINKED ISSUE #8: it takes 20 confirmations to shield funds
see: https://github.com/zcash/ZcashLightClientKit/issues/366

### LINKED ISSUE #9: Transparent funds are verified at Zero confs regardless
Transparent funds are verified at Zero confs regardless of what the min_confirmations are.

image

pacu:
Allow the caller to supply explicitly the number of confirmations required for TXOs when querying balances or sending funds.

Fixes #8
Fixes #9

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -766,13 +766,14 @@ pub extern "C" fn zcashlc_get_verified_balance(
     db_data_len: usize,
     account: i32,
     network_id: u32,
+    min_confirmations: u32,
 ) -> i64 {
     let res = catch_panic(|| {
         let network = parse_network(network_id)?;
         let db_data = wallet_db(db_data, db_data_len, network)?;
         if account >= 0 {
             (&db_data)
-                .get_target_and_anchor_heights(ANCHOR_OFFSET)
+                .get_target_and_anchor_heights(min_confirmations)
                 .map_err(|e| format_err!("Error while fetching anchor height: {}", e))
                 .and_then(|opt_anchor| {
                     opt_anchor
@@ -800,6 +801,7 @@ pub extern "C" fn zcashlc_get_verified_transparent_balance(
     db_data_len: usize,
     address: *const c_char,
     network_id: u32,
+    min_confirmations: u32,
 ) -> i64 {
     let res = catch_panic(|| {
         let network = parse_network(network_id)?;
@@ -807,7 +809,7 @@ pub extern "C" fn zcashlc_get_verified_transparent_balance(
         let addr = unsafe { CStr::from_ptr(address).to_str()? };
         let taddr = TransparentAddress::decode(&network, &addr).unwrap();
         let amount = (&db_data)
-            .get_target_and_anchor_heights(ANCHOR_OFFSET)
+            .get_target_and_anchor_heights(min_confirmations)
             .map_err(|e| format_err!("Error while fetching anchor height: {}", e))
             .and_then(|opt_anchor| {
                 opt_anchor
@@ -846,7 +848,7 @@ pub extern "C" fn zcashlc_get_total_transparent_balance(
         let addr = unsafe { CStr::from_ptr(address).to_str()? };
         let taddr = TransparentAddress::decode(&network, &addr).unwrap();
         let amount = (&db_data)
-            .get_target_and_anchor_heights(ANCHOR_OFFSET)
+            .get_target_and_anchor_heights(0u32)
             .map_err(|e| format_err!("Error while fetching anchor height: {}", e))
             .and_then(|opt_anchor| {
                 opt_anchor
@@ -1204,6 +1206,7 @@ pub extern "C" fn zcashlc_create_to_address(
     output_params: *const u8,
     output_params_len: usize,
     network_id: u32,
+    min_confirmations: u32,
 ) -> i64 {
     let res = catch_panic(|| {
         let network = parse_network(network_id)?;
@@ -1266,7 +1269,7 @@ pub extern "C" fn zcashlc_create_to_address(
             value,
             memo,
             OvkPolicy::Sender,
-            ANCHOR_OFFSET,
+            min_confirmations,
         )
         .map_err(|e| format_err!("Error while sending funds: {}", e))
     });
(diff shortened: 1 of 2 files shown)


rust/Cargo.toml
```diff
@@ -11,29 +11,21 @@ build = "build.rs"
...
 base58 = "0.1.0"
-#sha2 = "0.9"
-#bs58 = { version = "0.3", features = ["check"] }
-#hdwallet = "0.3.0"
-#ripemd160 = "0.9"
-hdwallet = { git = "https://github.com/nuttycom/hdwallet", rev = "72f1f7a56c114eed484cefd6d402b7ef28158712"}
-secp256k1 = "0.19"
+sha2 = "0.9"
+hdwallet = "0.3.0"
+ripemd160 = "0.9"
```
daira, line 23: The [`ripemd160` crate](https://crates.io/crates/ripemd160) is deprecated. Zcashd switched to using `ripemd` instead.

rust/src/lib.rs
```diff
@@ -289,52 +312,108 @@ pub unsafe extern "C" fn zcashlc_derive_extended_spending_keys(
...
+        extpub: encoded_pubkey.unwrap_or(std::ptr::null_mut()),
     }
 }
+use std::convert::TryInto;
+
+fn ufvk_from_ffi(
+    network: &Network,
+    ffiufvk: &FFIUnifiedViewingKey,
+) -> Option<UnifiedFullViewingKey> {
+    //TODO
```
daira, line 367: What is this TODO about? Please file an issue and link to it, or remove the TODO.

pacu: Thanks for pointing this out!

rust/src/lib.rs
```diff
@@ -299,42 +324,82 @@ pub struct FFIUVKBoxedSlice {
...
+        extfvk: encoded_extfvk.unwrap_or(std::ptr::null_mut()),
+        extpub: encoded_pubkey.unwrap_or(std::ptr::null_mut()),
     }
 }
+use std::convert::TryInto;
+
+fn ufvk_from_ffi(
+    network: &Network,
+    ffiufvk: &FFIUnifiedViewingKey,
+) -> Option<UnifiedFullViewingKey> {
```
nuttycom, line 352: Ideally this would return a `Result` that has better error information about how it failed.

rust/src/lib.rs
```diff
@@ -1375,33 +1518,45 @@ pub extern "C" fn zcashlc_shield_funds(
...
-            .map_err(|e| format_err!("Error while shielding transaction: {}", e))
+        shield_transparent_funds(
+            &mut update_ops,
+            &network,
+            LocalTxProver::new(spend_params, output_params),
+            &t_account_privkey,
+            &extfvk,
+            AccountId::from(account),
+            &memo_bytes,
+            ANCHOR_OFFSET,
```
str4d, line 1557: Dropping this here so I don't forget: I noticed that the Android SDK uses 0 here, which seems wrong (that would try and spend mempool funds).

rust/src/lib.rs
```diff
@@ -1268,25 +1392,27 @@ pub unsafe extern "C" fn zcashlc_derive_transparent_private_key_from_seed(
...
-        let index = if index >= 0 {
+        let _index = if index >= 0 {
             index as u32
         } else {
             return Err(format_err!("index argument must be positive"));
         };
-        let sk = derive_secret_key_from_seed(&network, &seed, AccountId(account), index).unwrap();
-        let sk_wif = Wif::from_secret_key(&sk, true);
+        
+        let account_priv_key = extended_private_key_from_seed(&network, seed, account);
```
nuttycom, line 1406: Would it make sense to derive the unified spending key here instead, and then extract the transparent part?

rust/src/lib.rs
```diff
@@ -1268,25 +1392,27 @@ pub unsafe extern "C" fn zcashlc_derive_transparent_private_key_from_seed(
...
+        
+        let encoded = serialize_extended_priv_key(&account_priv_key.unwrap());
 
-        Ok(CString::new(sk_wif.0.to_string()).unwrap().into_raw())
+        Ok(CString::new(hex::encode(encoded)).unwrap().into_raw())
```
nuttycom, line 1410: Since this changes the encoding, is there any risk that the old encoded version might have been persisted anywhere?

rust/src/lib.rs
```diff
@@ -1416,4 +1571,28 @@ fn parse_network(value: u32) -> Result<Network, failure::Error> {
...
+
+
+
+/// Serialization Helpers for AccountPrivKey 
+/// 
+/// 
+use zcash_primitives::consensus;
+use hdwallet::KeyIndex;
+use hdwallet::traits::Serialize;
+fn extended_private_key_from_seed<P: consensus::Parameters>(
```
nuttycom, line 1584: I'm a bit uncomfortable with duplicating this code, instead of just using the UnifiedSpendingKey::from_seed derivation.

nuttycom (approved):
utACK with suggestions.

nuttycom:
An additional suggestion: this should be `cargo fmt`'ed: #12

rust/src/lib.rs
pacu: The android SDK uses a version of the rust code that has a bug where funds took 10 confirmations or more. That's why you see a zero there. A problem that the iOS SDK also had and it's documented in many issues I think on the zcash-ios-wallet repo. Also another one of them is this one

https://github.com/zcash-hackworks/zcash-light-client-ffi/issues/8

rust/src/lib.rs
pacu: Solved!

rust/src/lib.rs
pacu: Oh, that's sound right. thanks

pacu: the problem I have with this is that I need to serialize this and I can't serialize an AccountPrivKey.

rust/src/lib.rs
pacu: Fortunately we don't persist it, but there's a good thing to document just in case anyone has.

rust/src/lib.rs
pacu: This can go away if there was a way serialize an AccountPrivKey since there's no API for doing so we need to fall back to ExtendedPrivKey and use that serialize function

nuttycom:
This has now been rebased atop #28

nuttycom (approved):
utACK

PR #28, opened 2022-06-27
Replace UnifiedViewingKey with UnifiedFullViewingKey

### LINKED ISSUE #15: Change to generating and using UnifiedFullViewingKeys
This issue only covers support for new wallets. Migrating existing wallets is covered by #18.

### LINKED ISSUE #16: Migrate to `zcash_primitives 0.6`

str4d:
UnifiedViewingKey had a bug that made it incompatible with ZIP 316.

Depends on https://github.com/zcash/librustzcash/pull/570.

Closes #15.
Closes #16.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.15.2"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e7a2e47a1fbe209ee101dd6d61285226744c6c8d3c21c8dc878ba6cb9f467f3a"
+checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
 dependencies = [
  "gimli",
 ]
@@ -18,34 +18,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "aead"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b613b8e1e3cf911a086f53f03bf286f52fd7a7258e4fa606f0ef220d39d8877"
+dependencies = [
+ "generic-array 0.14.5",
+]
+
+[[package]]
 name = "aes"
-version = "0.6.0"
+version = "0.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "884391ef1066acaa41e766ba8f596341b96e93ce34f9a43e7d24bf0a0eaf0561"
+checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
 dependencies = [
- "aes-soft",
- "aesni",
+ "cfg-if",
  "cipher",
-]
-
-[[package]]
-name = "aes-soft"
-version = "0.6.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "be14c7498ea50828a38d0e24a765ed2effe92a705885b57d029cd67d45744072"
-dependencies = [
- "cipher",
- "opaque-debug",
-]
-
-[[package]]
-name = "aesni"
-version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ea2e11f5e94c2f7d386164cc2aa1f97823fed6f259e486940a71c174dd01b0ce"
-dependencies = [
- "cipher",
- "opaque-debug",
+ "cpufeatures",
+ "opaque-debug 0.3.0",
 ]
 
 [[package]]
@@ -71,9 +61,9 @@ checksum = "a4c527152e37cf757a3f78aae5a06fbeefdb07ccc535c980a3208ee3060dd544"
 
 [[package]]
 name = "arrayvec"
-version = "0.5.2"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
+checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "atty"
@@ -94,13 +84,13 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "backtrace"
-version = "0.3.59"
+version = "0.3.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4717cfcbfaa661a0fd48f8453951837ae7e8f81e481fbb136e3202d72805a744"
+checksum = "cab84319d616cfb654d03394f38ab7e6f0919e181b1b57e1fd15e7fb4077d9a7"
 dependencies = [
  "addr2line",
  "cc",
- "cfg-if 1.0.0",
+ "cfg-if",
  "libc",
  "miniz_oxide",
  "object",
@@ -109,9 +99,9 @@ dependencies = [
 
 [[package]]
 name = "base-x"
-version = "0.2.10"
+version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc19a4937b4fbd3fe3379793130e42060d10627a360f2127802b10b87e7baf74"
+checksum = "4cbbc9d0964165b47557570cce6c952866c2678457aca742aafc9fb771d30270"
 
 [[package]]
 name = "base58"
@@ -126,6 +116,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
+name = "base64ct"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a32fd6af2b5827bce66c29053ba0e7c42b9dcab01835835058558c10851a46b"
+
+[[package]]
 name = "bech32"
 version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -133,22 +129,37 @@ checksum = "cf9ff0bbfd639f15c74af777d81383cf53efb7c93613f6cab67c6c11e05bbf8b"
 
 [[package]]
 name = "bellman"
-version = "0.8.1"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7089887635778eabf0038a166f586eee5413fb85c8fa6c9a754914f0f644f49f"
+checksum = "a4dd656ef4fdf7debb6d87d4dd92642fcbcdb78cbf6600c13e25c87e4d1a3807"
 dependencies = [
- "bitvec 0.18.5",
+ "bitvec",
  "blake2s_simd",
  "byteorder",
- "crossbeam",
+ "crossbeam-channel",
  "ff",
- "futures",
- "futures-cpupool",
  "group",
+ "lazy_static",
+ "log",
  "num_cpus",
  "pairing",
- "rand_core 0.5.1",
+ "rand_core",
+ "rayon",
  "subtle",
+]
+
+[[package]]
+name = "bip0039"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d0830ae4cc96b0617cc912970c2b17e89456fecbf55e8eed53a956f37ab50c41"
+dependencies = [
+ "hmac",
+ "pbkdf2",
+ "rand",
+ "sha2 0.9.9",
+ "unicode-normalization",
+ "zeroize",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


rust/src/lib.rs
```diff
@@ -460,42 +442,36 @@ pub unsafe extern "C" fn zcashlc_derive_transparent_address_from_public_key(
...
-    
+        let taddr = legacy::keys::pubkey_to_address(&pk).encode(&network);
+
         Ok(CString::new(taddr).unwrap().into_raw())
     });
     unwrap_exc_or_null(res)
 }
 
-/// derives a shielded address from the given viewing key. 
+/// derives a shielded address from the given viewing key.
```
pacu, line 452:
```suggestion
/// derives a shielded address from the given unified viewing key.
```

pacu (changes requested):
I left some questions and comments.

Great progress!

nuttycom (changes requested):
utACK modulo updating the changelog.

rust/Cargo.lock
```diff
@@ -133,58 +129,62 @@ checksum = "cf9ff0bbfd639f15c74af777d81383cf53efb7c93613f6cab67c6c11e05bbf8b"
...
+name = "bip0039"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+checksum = "d0830ae4cc96b0617cc912970c2b17e89456fecbf55e8eed53a956f37ab50c41"
+dependencies = [
+ "hmac",
+ "pbkdf2",
+ "rand",
+ "sha2 0.9.9",
```
daira, line 160: zcash_primitives, bs58 and bip0039 could ideally do with updating to use the newer sha2 0.10.2, but it's not critical.

rust/Cargo.lock
```diff
@@ -1622,53 +1987,78 @@ dependencies = [
...
+[[package]]
+name = "zcash_note_encryption"
+version = "0.1.0"
+source = "git+https://github.com/zcash/librustzcash.git?rev=37fc28634e811fdf8db9274a0080cff17c6b6a09#37fc28634e811fdf8db9274a0080cff17c6b6a09"
+dependencies = [
+ "chacha20",
+ "chacha20poly1305",
+ "rand_core",
  "subtle",
 ]
```
daira, line 2019: This duplication seems undesirable.

rust/Cargo.lock
```diff
@@ -1062,15 +1319,24 @@ dependencies = [
...
+name = "ripemd"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1facec54cb5e0dc08553501fa740091086d0259ad0067e0d4103448e4cb22ed3"
+dependencies = [
+ "digest 0.10.3",
+]
+
 [[package]]
 name = "ripemd160"
```
daira, line 1332: Ideally we would only depend on ripemd. The dependency on ripemd160 is from hdwallet-bitcoin.

rust/src/lib.rs
```diff
@@ -153,28 +158,31 @@ pub extern "C" fn zcashlc_init_accounts_table(
...
+        let usks: Vec<_> = (0..accounts)
             .map(|account| {
-                let tsk =
-                    derive_secret_key_from_seed(&network, &seed, AccountId(account), 0).unwrap();
-                derive_transparent_address_from_secret_key(&tsk)
+                let account_id = AccountId::from(account);
+                UnifiedSpendingKey::from_seed(&network, &seed, account_id)
+                    .map(|usk| (account_id, usk))
+                    .map_err(
+                    |e| format_err!("error generating unified spending key from seed: {:?}", e),
```
daira, line 167: This doesn't look rustfmt'd.

rust/src/lib.rs
```diff
@@ -462,29 +460,25 @@ pub unsafe extern "C" fn zcashlc_derive_transparent_address_from_public_key(
 /// call zcashlc_string_free with the returned pointer when done using it
 #[no_mangle]
 pub unsafe extern "C" fn zcashlc_derive_shielded_address_from_viewing_key(
-    extfvk: *const c_char,
+    ufvk: *const c_char,
```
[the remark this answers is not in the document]
daira: Maybe change the name to `zcashlc_derive_shielded_address_from_unified_viewing_key` so that callers are forced to change?

rust/Cargo.lock
nuttycom: That's weird, I wonder how that got introduced.

rust/src/lib.rs
```diff
@@ -720,7 +714,7 @@ pub extern "C" fn zcashlc_get_verified_balance(
         let db_data = wallet_db(db_data, db_data_len, network)?;
         if account >= 0 {
             (&db_data)
-                .get_target_and_anchor_heights()
+                .get_target_and_anchor_heights(ANCHOR_OFFSET)
```
daira, line 717: ANCHOR_OFFSET is defined as 10. Is this the same as the previous behaviour?

rust/src/lib.rs
```diff
@@ -764,14 +758,15 @@ pub extern "C" fn zcashlc_get_verified_transparent_balance(
...
                     .map_err(|e| {
                         format_err!("Error while fetching verified transparent balance: {}", e)
                     })
             })?
             .iter()
-            .map(|utxo| utxo.value)
-            .sum::<Amount>();
+            .map(|utxo| utxo.txout.value)
+            .sum::<Option<Amount>>()
+            .ok_or_else(|| format_err!("Balance overflowed MAX_MONEY. Lucky you!"))?;
```
daira, line 769:
```suggestion
            .ok_or_else(|| format_err!("Balance overflowed MAX_MONEY"))?;
```

rust/src/lib.rs
daira: Ah you said below that it is, @nuttycom.

nuttycom: Yes. It was hardcoded inside librustzcash.

rust/Cargo.lock
nuttycom: Oh, this is just because of the patch dependency.

rust/src/lib.rs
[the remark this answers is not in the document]
pacu: I'm not aware of any view-only sapling wallets that use the SDK, but it will be problematic if there was. It's probably safe to do that change and respond to any issues filed about this API changes if so.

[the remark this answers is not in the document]
pacu: Something it could be useful to keep around is sapling Spending Key helpers. If you have a guarda wallet, the current API can't support moving those funds out because it used to back up a sapling spending key instead of a seed phrase. 

Again, an edge case.

rust/src/lib.rs
```diff
@@ -755,7 +753,7 @@ pub extern "C" fn zcashlc_get_verified_transparent_balance(
         let addr = unsafe { CStr::from_ptr(address).to_str()? };
         let taddr = TransparentAddress::decode(&network, &addr).unwrap();
         let amount = (&db_data)
-            .get_target_and_anchor_heights()
+            .get_target_and_anchor_heights(ANCHOR_OFFSET)
```
pacu, line 756: we should double check that this aligns with the wallet balance that `shield_funds` uses. There's currently a Bug of `shield_funds` requiring 20 confirmations instead of 10.

rust/src/lib.rs
```diff
@@ -1306,30 +1340,48 @@ pub unsafe extern "C" fn zcashlc_derive_transparent_address_from_seed(
...
     unwrap_exc_or_null(res)
 }
 
-/// Derives a transparent address from the given secret key enconded as a WIF string
+/// Derives a transparent address from the given account private key.
 #[no_mangle]
-pub unsafe extern "C" fn zcashlc_derive_transparent_address_from_secret_key(
-    tsk: *const c_char,
+pub unsafe extern "C" fn zcashlc_derive_transparent_address_from_account_private_key(
+    xprv: *const c_char,
```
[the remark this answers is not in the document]
pacu: This is only used on `shield_funds` 
unless the wallet is storing this private key, we should refactor the SDK's public API to require  USK for shielding funds and this could be removed and replaced with migration documentation

rust/src/lib.rs
```diff
@@ -1340,7 +1396,7 @@ pub extern "C" fn zcashlc_shield_funds(
     db_data: *const u8,
     db_data_len: usize,
     account: i32,
-    tsk: *const c_char,
+    xprv: *const c_char,
```
pacu, line 1399: can UnifiedSpendingKeys be encoded on a string representation we could pass to this method instead of extsk and xpriv?

pacu: I guess that's for another PR 😬

pacu (approved):
LGTM. Left Non-blocking comments

rust/src/lib.rs
nuttycom: We have intentionally not defined a USK string representation.

str4d:
[Force-pushed](https://github.com/zcash-hackworks/zcash-light-client-ffi/compare/a4f3b38c48c95b3d4400b3002a230008f67d68e7..d34183801a91d247ea602ce0de00993001fa1f9d) to make some minor cleanups, split out part of the changes into a separate commit, add a missing FFI method, and rename another.

The first commit of this PR should now be (semantically) identical to the changes made to the Android SDK in https://github.com/zcash/zcash-android-wallet-sdk/pull/530, except for two things:
- This repo has a `zcashlc_init_accounts_table` FFI method that the Android SDK does not have.
- @nuttycom altered this PR to use a slightly newer revision of `zcash/librustzcash`, which affects the `zcashlc_init_accounts_table*` FFI methods.

rust/Cargo.lock
nuttycom: The ripemd dependency has now been removed entirely, as it's transitive.

daira (commented):
Flushing comments

rust/src/lib.rs
```diff
@@ -1306,30 +1344,48 @@ pub unsafe extern "C" fn zcashlc_derive_transparent_address_from_seed(
...
+        };
+        let taddr = taddr.encode(&network);
 
         Ok(CString::new(taddr).unwrap().into_raw())
     });
     unwrap_exc_or_null(res)
 }
 
-/// Derives a transparent address from the given secret key enconded as a WIF string
+/// Derives a transparent address from the given account private key.
```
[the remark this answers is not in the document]
str4d: We don't refer to these as "extended private keys" in the ZIP 316 context AFAIR; "account private key" is effectively a newtype in this context.

str4d:
[Force-pushed](https://github.com/zcash-hackworks/zcash-light-client-ffi/compare/d34183801a91d247ea602ce0de00993001fa1f9d..afb9dbd2194ebdb38844302b19cc7ac06b192d00) to fix an FFI method name and add another type-checking FFI method for UFVKs.

rust/src/lib.rs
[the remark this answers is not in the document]
nuttycom: Ideally we'd support that use case with a more comprehensive approach to spending key import. It could still be satisfied using USKs, using a Sapling-only USK.

rust/src/lib.rs
```diff
@@ -1248,41 +1304,68 @@ pub extern "C" fn zcashlc_vec_string_free(v: *mut *mut c_char, len: usize, capac
...
+        // Encode using the BIP 32 xprv serialization format.
+        let xprv_str: String = xprv.serialize();
 
-        Ok(CString::new(sk_wif.0.to_string()).unwrap().into_raw())
+        Ok(CString::new(xprv_str).unwrap().into_raw())
     });
     unwrap_exc_or_null(res)
 }
 
+fn p2pkh_xprv<P: Parameters>(
```
nuttycom, line 1345: Why is this not using https://github.com/zcash/librustzcash/blob/main/zcash_primitives/src/legacy/keys.rs#L24 ?

rust/src/lib.rs
```diff
@@ -1032,7 +1024,6 @@ pub extern "C" fn zcashlc_scan_blocks(
 pub extern "C" fn zcashlc_put_utxo(
     db_data: *const u8,
     db_data_len: usize,
-    address_str: *const c_char,
```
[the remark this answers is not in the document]
pacu: How shall we clear them up since the SDK doesn't account for the Script bytes from an account perspective?

rust/src/lib.rs
str4d: I needed specifically the `hdwallet_bitcoin::PrivKey` type here, and while that does store an `hdwallet::ExtendedPrivKey` internally, and the helper you're referencing returns a newtype around `hdwallet::ExtendedPrivKey`, I don't have any way to access it (and I'm not sure I want one in the public API of the Rust crates just yet). What we really need is serialization APIs on `AccountPrivKey`, which could then replace this logic.

nuttycom: The main thing that I don't care for is having the derivation path logic reimplemented, but for now I guess that's okay.

str4d:
[Force-pushed](https://github.com/zcash-hackworks/zcash-light-client-ffi/compare/222896736c81023f5c469b9e18594553ea6a25c1..afb9dbd2194ebdb38844302b19cc7ac06b192d00) to remove the `make xcframework` commit (which we will instead add to the feature branch once we are done constructing it).

rust/src/lib.rs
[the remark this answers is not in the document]
str4d: I'm not sure what you mean; from the perspective of a single UTXO, you just have an address it was sent to. The existing `put_utxo` FFI method doesn't care about accounts at all.

rust/Cargo.lock
str4d: Correct, not critical and not going to happen in this PR.

rust/src/lib.rs
```diff
@@ -462,29 +464,25 @@ pub unsafe extern "C" fn zcashlc_derive_transparent_address_from_public_key(
 /// call zcashlc_string_free with the returned pointer when done using it
```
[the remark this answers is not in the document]
str4d: There are a bunch of cleanups that need doing like this, which can be done in a separate PR.

PR #31, opened 2022-08-21
Remove Spending key from `shield_funds`

pacu:
Remove Spending key from `shield_funds` and use Sapling extended full viewing key instead.

This is moving work forward to ultimately use UnifiedFullViewingKey from the db and only requiring account priv key as a parameter

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -37,10 +37,10 @@ lto = true
 # We also need zcash_client_backend and zcash_client_sqlite, which haven't been published
 # with NU5 updates yet.
 [patch.crates-io]
-zcash_client_backend = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
-zcash_client_sqlite = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
-zcash_primitives = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
-zcash_proofs = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
+zcash_client_backend = { git = 'https://github.com/zcash/librustzcash.git', rev='8c00ca3b88d0fea880ccd731544ad312caa0e6af' }
+zcash_client_sqlite = { git = 'https://github.com/zcash/librustzcash.git', rev='8c00ca3b88d0fea880ccd731544ad312caa0e6af' }
+zcash_primitives = { git = 'https://github.com/zcash/librustzcash.git', rev='8c00ca3b88d0fea880ccd731544ad312caa0e6af' }
+zcash_proofs = { git = 'https://github.com/zcash/librustzcash.git', rev='8c00ca3b88d0fea880ccd731544ad312caa0e6af' }
 
 [features]
 mainnet = ["zcash_client_sqlite/mainnet"]
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -1809,8 +1809,6 @@ pub extern "C" fn zcashlc_derive_transparent_address_from_account_private_key(
 ///   documentation of pointer::offset.
 /// - `xprv` must be non-null and must point to a null-terminated UTF-8 string representing
 ///   a Base58-encoded transparent spending key.
-/// - `extsk` must be non-null and must point to a null-terminated UTF-8 string representing
-///   a Bech32-encoded Sapling extended spending key for the given network.
 /// - `spend_params` must be non-null and valid for reads for `spend_params_len` bytes, and it must have an
 ///   alignment of `1`. Its contents must be the Sapling spend proving parameters.
 /// - The memory referenced by `spend_params` must not be mutated for the duration of the function call.
@@ -1827,7 +1825,6 @@ pub extern "C" fn zcashlc_shield_funds(
     db_data_len: usize,
     account: i32,
     xprv: *const c_char,
-    extsk: *const c_char,
     memo: *const c_char,
     spend_params: *const u8,
     spend_params_len: usize,
@@ -1848,7 +1845,6 @@ pub extern "C" fn zcashlc_shield_funds(
             return Err(format_err!("account argument must be positive"));
         };
         let xprv_str = unsafe { CStr::from_ptr(xprv) }.to_str()?;
-        let extsk = unsafe { CStr::from_ptr(extsk) }.to_str()?;
         let memo = unsafe { CStr::from_ptr(memo) }.to_str()?;
         let spend_params = Path::new(OsStr::from_bytes(unsafe {
             slice::from_raw_parts(spend_params, spend_params_len)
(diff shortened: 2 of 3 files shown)


nuttycom (approved):
utACK

pacu:
There are a couple of things that I'm still figuring out. It's possible that we need derivation from extfvk to sapling address back in our derivation API

rust/src/lib.rs
```diff
@@ -539,6 +539,36 @@ pub unsafe extern "C" fn zcashlc_derive_extended_full_viewing_key(
     unwrap_exc_or_null(res)
 }
 
+/// derives a shielded address from the given viewing key.
+/// call zcashlc_string_free with the returned pointer when done using it
+#[no_mangle]
+pub unsafe extern "C" fn zcashlc_derive_shielded_address_from_viewing_key(
```
nuttycom, line 545:
```suggestion
pub unsafe extern "C" fn zcashlc_derive_sapling_address_from_viewing_key(
```

nuttycom (commented):
We should not use the term `shielded` in API methods where we mean `sapling`.

str4d (changes requested):
Marking that I want to review this.

rust/src/lib.rs
```diff
@@ -712,6 +712,36 @@ pub extern "C" fn zcashlc_derive_extended_full_viewing_key(
...
+///
+/// # Safety
+///
+/// - `extfvk` must be non-null and must point to a null-terminated UTF-8 string representing
+///   a Bech32-encoded Sapling extended full viewing key for the given network.
+/// - The memory referenced by `extfvk` must not be mutated for the duration of the function call.
+/// - Call [`zcashlc_string_free`] to free the memory associated with the returned pointer
+///   when done using it.
+#[no_mangle]
+pub extern "C" fn zcashlc_derive_sapling_address_from_viewing_key(
```
str4d, line 725: Where or why is this needed? I don't think we should be exposing Sapling addresses after this SDK upgrade (the Sapling addresses are being merged into Unified Addresses), but if there's a use case for this I'd like to know.

rust/src/lib.rs
```diff
@@ -1875,7 +1875,7 @@ pub extern "C" fn zcashlc_shield_funds(
     db_data_len: usize,
     account: i32,
     xprv: *const c_char,
-    extsk: *const c_char,
+    extfvk: *const c_char,
```
str4d, line 1932: I think we could probably just remove this field entirely. `zcashlc_shield_funds` is already being used (AFAIK) uniformly with an `extsk` that corresponds to the given `account`, so we should just remove the `extsk / extfvk` parameter entirely and look up the UFVK from `db_data` (and then send to the Sapling component of it).

rust/src/lib.rs
nuttycom: I think that this an intermediate step to get the current front-end wallet code working with current librustzcash.

rust/src/lib.rs
nuttycom: How does this work with Edge, if they insist on managing all their own keys? is the `accounts` table even populated in their implementation?

rust/src/lib.rs
pacu: This is mostly for a validation done in the shield funds function on swift to make sure that the recipient of this transaction is "self". Ideally this will go entirely but being that the shield_funds function on rust crate (not the ffi) requires a extfvk we are forced to maintain this until we change that. 

Also. we can review if such validation is needed and decide whether ir should be done or not

rust/src/lib.rs
str4d: They manage the _spending keys_, not the _viewing keys_. Viewing keys have always been populated in the `accounts` table for every SDK user.

rust/src/lib.rs
str4d: Oh, I see what you mean: we can't get an `extfvk` format out of a UFVK because we don't store everything needed. In that case, we can leave the `extfvk` argument to `zcashlc_shield_funds` and I'll fix that in a subsequent PR.

I still don't think that this method is necessary; the SDK is already trusting the Rust side to do everything else, so why wouldn't it trust the Rust side to correctly shield the funds to an address underneath `account`?

rust/src/lib.rs
nuttycom: I'm actually changing this in librustzcash now; we need to be shielding to the internal (change) address anyway. So then I'll propagate that change outward. And maybe we won't need zcashlc_derive_sapling_address_from_viewing_key then anyway.

str4d: Opened https://github.com/zcash/librustzcash/issues/614 for removing the `extfvk` argument.

pacu: > I think we could probably just remove this field entirely. `zcashlc_shield_funds` is already being used (AFAIK) uniformly with an `extsk` that corresponds to the given `account`, so we should just remove the `extsk / extfvk` parameter entirely and look up the UFVK from `db_data` (and then send to the Sapling component of it).

Yes, we discussed this internally and I think is a positive change. Is less error prone and more coherent. Otherwise if there's a developer error from SDK clients, they could turn the `shield_funds` into some weird `send_transparent_funds_somewhere`

pacu: also we can control where the shielding is done.

nuttycom: https://github.com/zcash/librustzcash/pull/615

rust/Cargo.toml
```diff
@@ -37,10 +37,10 @@ lto = true
...
 # with NU5 updates yet.
 [patch.crates-io]
-zcash_client_backend = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
-zcash_client_sqlite = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
-zcash_primitives = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
-zcash_proofs = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
+zcash_client_backend = { git = 'https://github.com/nuttycom/librustzcash.git', rev='82a71a024a30530a0d5734f83079389ef0004ae6' }
+zcash_client_sqlite = { git = 'https://github.com/nuttycom/librustzcash.git', rev='82a71a024a30530a0d5734f83079389ef0004ae6' }
+zcash_primitives = { git = 'https://github.com/nuttycom/librustzcash.git', rev='82a71a024a30530a0d5734f83079389ef0004ae6' }
+zcash_proofs = { git = 'https://github.com/nuttycom/librustzcash.git', rev='82a71a024a30530a0d5734f83079389ef0004ae6' }
```
daira, line 43: I have not checked this hash.

rust/src/lib.rs
nuttycom: I've removed this commit, pursuant to zcash/librustzcash#615 merging.

daira (approved):
utACK from me.

rust/Cargo.toml
```diff
@@ -37,10 +37,10 @@ lto = true
...
 # with NU5 updates yet.
 [patch.crates-io]
-zcash_client_backend = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
-zcash_client_sqlite = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
-zcash_primitives = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
-zcash_proofs = { git = 'https://github.com/zcash/librustzcash.git', rev='f800f9f1446850e5009f88a16ed84ea19d573faf' }
+zcash_client_backend = { git = 'https://github.com/zcash/librustzcash.git', rev='8c00ca3b88d0fea880ccd731544ad312caa0e6af' }
+zcash_client_sqlite = { git = 'https://github.com/zcash/librustzcash.git', rev='8c00ca3b88d0fea880ccd731544ad312caa0e6af' }
+zcash_primitives = { git = 'https://github.com/zcash/librustzcash.git', rev='8c00ca3b88d0fea880ccd731544ad312caa0e6af' }
+zcash_proofs = { git = 'https://github.com/zcash/librustzcash.git', rev='8c00ca3b88d0fea880ccd731544ad312caa0e6af' }
```
str4d, line 43: I checked this commit hash.

str4d (approved):
utACK d7d2ab93296ba881207b90ea49b7a058298a5012

PR #43, opened 2022-09-07
Fix get_{received|sent}_memo functions to use u8* 

pacu:


nuttycom:
Closing in favor of #44 which eliminates some unnecessarily duplicated code.

PR #46, opened 2022-09-09
Add FFI method to access transparent receiver of UA

str4d:


pacu (approved):

rust/src/lib.rs
```diff
@@ -790,6 +792,62 @@ pub extern "C" fn zcashlc_get_address(
...
+                TransparentAddress::Script(data) => {
+                    ZcashAddress::from_transparent_p2sh(network, *data)
+                }
+            };
+
+            Ok(CString::new(taddr.encode()).unwrap().into_raw())
+        } else {
+            Err(format_err!(
+                "Unified Address doesn't contain a transparent receiver"
+            ))
```
str4d, line 845: @pacu would you prefer that we return an error if there is no transparent receiver, or just `Ok(null)` (so checking the last error would show nothing, indicating it just didn't have a transparent receiver)? Given that I believe you only plan to call this on UAs generated by the SDK, we can assume a transparent receiver is always intended to be present if this is being called.

pacu: I think the error is fine.

rust/src/lib.rs
```diff
@@ -790,6 +792,62 @@ pub extern "C" fn zcashlc_get_address(
     unwrap_exc_or_null(res)
 }
 
+struct UnifiedAddressParser(UnifiedAddress);
+
+impl zcash_address::TryFromRawAddress for UnifiedAddressParser {
```
str4d, line 797: This exists because we don't have this implemented on `UnifiedAddress` itself (only on `ReceiverAddress`), and it's faster to just do this to unblock the PR. We can move this into `zcash_client_backend` at some point.

rust/src/lib.rs
```diff
@@ -790,6 +792,62 @@ pub extern "C" fn zcashlc_get_address(
     unwrap_exc_or_null(res)
 }
 
+struct UnifiedAddressParser(UnifiedAddress);
```
nuttycom, line 795: nit: this is weirdly named, because this isn't a parser. But it's internal-only, so I guess it's fine.

rust/src/lib.rs
nuttycom: An alternative is to use the parsing code `PaymentAddress` and then just pattern match on the result. That would be a tiny bit more flexible, as it'd work both with taddr strings and UA strings.

nuttycom (approved):
utACK, though I think that using the `PaymentAddress` parser would perhaps be a better; that would avoid the need to add the zcash_address dependency.

rust/src/lib.rs
str4d: I assume you mean the `ReceiverAddress` parsing code (as `PaymentAddress` is specifically Sapling). But we don't want this to work for taddr strings? For those it would be a no-op, and a success there would mislead the caller, who is expecting to receive a component of a Unified Address.

daira (approved):
utACK

rust/src/lib.rs
str4d: It technically is, as the `try_into()` is parsing the internal receivers of the `zcash_address::kind::unified::Address` to obtain a valid `UnifiedAddress`.

rust/src/lib.rs
nuttycom: Ah, actually I meant `RecipientAddress`; `ReceiverAddress` isn't a thing and I was confused. Anyway, one could still return an error for taddr strings if you wanted to, and it'd avoid the `zcash-address` dependency.

PR #47, opened 2022-09-09
Add FFI method to extract typecodes of a UA's receivers

str4d:


rust/src/lib.rs
```diff
@@ -792,6 +796,64 @@ pub extern "C" fn zcashlc_get_address(
...
+/// slice.
+///
+/// # Safety
+///
+/// - `ua` must be non-null and must point to a null-terminated UTF-8 string containing an
+///   encoded Unified Address.
+/// - Call [`zcashlc_free_typecodes`] to free the memory associated with the returned
+///   pointer when done using it.
+#[no_mangle]
+pub extern "C" fn zcashlc_get_typecodes_for_unified_address_receivers(
```
pacu, line 811: we should document how the caller should map the u32's to types

pacu (changes requested):

rust/src/lib.rs
str4d: Done.

pacu (approved):

rust/src/lib.rs
```diff
@@ -792,6 +796,69 @@ pub extern "C" fn zcashlc_get_address(
...
+/// See the following sections of ZIP 316 for details on how to interpret typecodes:
+/// - [List of known typecodes](https://zips.z.cash/zip-0316#encoding-of-unified-addresses)
+/// - [Adding new types](https://zips.z.cash/zip-0316#adding-new-types)
+/// - [Metadata Items](https://zips.z.cash/zip-0316#metadata-items)
+///
+/// # Safety
+///
+/// - `ua` must be non-null and must point to a null-terminated UTF-8 string containing an
+///   encoded Unified Address.
+/// - Call [`zcashlc_free_typecodes`] to free the memory associated with the returned
```
nuttycom, line 813: It would be nice for us to be consistent about how we return slices; #45 creates an explicit struct to return both the pointer and the length, though that will hopefully become unnecessary if we change how account table initialization is done; in that case, this will be the only operation that is returning a slice, at which point it will be consistent with itself and changing it can wait, since hopefully all of this will be disappearing in favor of uniffi anyway.

nuttycom (approved):
utACK with comment

PR #66, opened 2022-11-16
Update to librustzcash version 0.7, take 2

### LINKED ISSUE #65: Enable the `deny(unsafe_op_in_unsafe_fn)` lint everywhere in zcash-light-client-ffi we define `unsafe` functions
From https://github.com/zcash-hackworks/zcash-light-client-ffi/pull/32/files#r1022962018 :

A side-effect of declaring a function `unsafe` is that unsafe operations in its body are no longer required to be in `unsafe` blocks (see [Rust RFC 2585](https://rust-lang.github.io/rfcs/2585-unsafe-block-in-unsafe-fn.html) for a proposal to change this, not yet adopted). Ideally, we should enable the `deny(unsafe_op_in_unsafe_fn)` lint everywhere we define `unsafe` functions, since the style we write FFI functions in is mostly already consistent with that lint.

sellout:
First time I failed to include the ZIP 317 changes in the feature branch.
closes #65

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -1524,6 +1524,18 @@ pub extern "C" fn zcashlc_get_received_memo(
     )
 }
 
+/// Returns the memo for a note by copying the corresponding bytes to the received
+/// pointer in `memo_bytes_ret`.
+///
+/// # Safety
+///
+/// - `db_data` must be non-null and valid for reads for `db_data_len` bytes, and it must have an
+///   alignment of `1`. Its contents must be a string representing a valid system path in the
+///   operating system's preferred representation.
+/// - The memory referenced by `db_data` must not be mutated for the duration of the function call.
+/// - The total size `db_data_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of pointer::offset.
+/// - `memo_bytes_ret` must be non-null and must point to an allocated 512-byte region of memory.
 fn zcashlc_get_memo(
     db_data: *const u8,
     db_data_len: usize,
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -1,3 +1,5 @@
+#![deny(unsafe_op_in_unsafe_fn)]
+
 use failure::format_err;
 use ffi_helpers::panic::catch_panic;
 use schemer::MigratorError;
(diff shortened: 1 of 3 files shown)


sellout:
These changes are identical to #63. That PR has approval from @daira, @pacu, and @str4d. So if one of you could approve this with a message that says it’s fine to carry over those approvals, I can merge this quickly.

rust/src/lib.rs
```diff
@@ -2177,16 +2215,41 @@ pub extern "C" fn zcashlc_shield_funds(
...
-            &network,
-            LocalTxProver::new(spend_params, output_params),
-            &usk,
-            &taddrs,
-            &memo_bytes,
-            ANCHOR_OFFSET,
-        )
-        .map_err(|e| format_err!("Error while shielding transaction: {}", e))
+        if use_zip317_fees {
+            let input_selector = GreedyInputSelector::new(
```
pacu, line 2219: can this be refactored to 
````
let input_selector  = if  use_zip317_fees {
 // create ZIP-317 selector
 } else {
  // create ZIP-313 selector
  }
  
// single call to 
   shield_transparent_funds(
                &mut update_ops,
                &network,
                LocalTxProver::new(spend_params, output_params),
                &input_selector,
                &usk,
                &taddrs,
                &memo_bytes,
                ANCHOR_OFFSET,
            )
            .map_err(|e| format_err!("Error while shielding transaction: {}", e))
            
````

pacu (changes requested):
can we remove this repeated code?

rust/src/lib.rs
str4d: Sadly no, because the two branches of the conditional have different types. In the equivalent PR on the Android SDK, I squared the circle (and also deduplicated the logic between `createToAddress` and `shieldToAddress`) with this helper: https://github.com/zcash/zcash-android-wallet-sdk/pull/775/files#diff-ea14a0f1195cb3440fbdf0cfba825a92078a0d1ff9655232f182000d360e2976R1069

daira:
https://github.com/zcash-hackworks/zcash-light-client-ffi/pull/67 fixes issues with `unsafe` annotations (and #65).

daira (approved):
utACK. (Consider including #67, but it's fine to do that separately.)

daira (approved):
Re-review after merges: utACK d305bb7e87fc19c4e0d033c214fbc156b81196e6

pacu (approved):
Tested autoshielding confirmations seem to be fine.

PR #67, opened 2022-11-16
`unsafe` annotations

### LINKED ISSUE #65: Enable the `deny(unsafe_op_in_unsafe_fn)` lint everywhere in zcash-light-client-ffi we define `unsafe` functions

daira:
This enables `deny(unsafe_op_in_unsafe_fn)` globally and adds `unsafe` annotations as necessary. fixes #65

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -163,7 +163,7 @@ pub extern "C" fn zcashlc_clear_last_error() {
 /// - The total size `seed_len` must be no larger than `isize::MAX`. See the safety documentation
 ///   of pointer::offset.
 #[no_mangle]
-pub extern "C" fn zcashlc_init_data_database(
+pub unsafe extern "C" fn zcashlc_init_data_database(
     db_data: *const u8,
     db_data_len: usize,
     seed: *const u8,
@@ -264,7 +264,7 @@ pub unsafe extern "C" fn zcashlc_free_binary_key(ptr: *mut FFIBinaryKey) {
 ///
 /// [ZIP 316]: https://zips.z.cash/zip-0316
 #[no_mangle]
-pub extern "C" fn zcashlc_create_account(
+pub unsafe extern "C" fn zcashlc_create_account(
     db_data: *const u8,
     db_data_len: usize,
     seed: *const u8,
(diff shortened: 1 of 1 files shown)


rust/.cargo/config
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["-Dunsafe_op_in_unsafe_fn"]
```
str4d, line 2: Why this instead of adding `#![deny(unsafe_op_in_unsafe_fn)]` top the top of `rust/src/lib.rs`?

str4d (changes requested):
PR changes LGTM; only issue is how the lint is enabled.

I confirmed that the lint was stabilised in 1.51.0, and our MSRV for the Zcash Rust crates is currently 1.56.1, so we can use this lint everywhere.

rust/.cargo/config
sellout: I like it here, since it enables it for any new files that may be added.

daira: Since `rust/src/lib.rs` is the only source file at the moment, they are equivalent. This approach is arguably better for multi-file projects (like librustzcash) if our policy is going to be to enable this lint in all of our code. So I decided to do it in the same way here as I would do it in librustzcash.

sellout (approved):
utACK

Should we also add a `rust-version = "1.51"` (or match the MSRV) line to Cargo.toml to catch anyone trying to build with an older version?

rust/.cargo/config
str4d: Lints applied at the module level apply to every module under the module they are applied on&lang;/wordsalad&rang;. Using `#![deny(LINT)]` in `lib.rs` makes it a crate-wide lint, which for a single crate is equivalent to passing a compiler flag. And I generally prefer placing things like this into the source code when possible, instead of adding compiler flags (which need more care when interacting with the rest of the build system, particularly in multi-language codebases like this one).

str4d: For multi-crate workspaces with many lints, having to duplicate across `lib.rs` files can become cumbersome, which is something being discussed and worked on in https://github.com/rust-lang/cargo/issues/5034. But that argument does not apply here as this is (and will always be) a single-crate workspace.

str4d (approved):

str4d:
> Should we also add a `rust-version = "1.51"` (or match the MSRV) line to Cargo.toml to catch anyone trying to build with an older version?

Yes at some point (I have already done this in the Android SDK), but doesn't have to happen in this PR (as the MSRV is unaltered by it). The actual MSRV is at least 1.56.1 (and is 1.59 over in the Android SDK, but I think that's because of an Android-specific dependency).

PR #77, opened 2022-12-13
Implement FsBlockDb

### LINKED ISSUE #73: `shield_funds` with ZIP-313 fees can't shield less than 100000 zatoshi.
See ZcashLightClientKit `testShieldFunds()` 

When running that test, the wallet attempts to shield 10000 zats but it fails where when using fixed ZIP-313 fees when it shouldn't 
````

-[DarksideTests.ShieldFundsTests testShieldFunds] : failed - Failed With error: RustWeldingError generic error: Error while shielding transaction: Insufficient balance (have 10000, need 100000 including fee)

````

### LINKED ISSUE #81: Adopt latest rust crates

pacu:
This is the initial implementation of the rust FFI code for using FsBlockDb from Swift. 

It's based on https://github.com/zcash/zcash-android-wallet-sdk/pull/804

It points to a branch because I need to code a fmt::Display trait compliance for the FsBlockError enum which I could implement from the FFI module itself. 
Closes #81 
Closes #73

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 FsBlock Db implementation and removal of BlockBb cache.
 
-Implement `zcashlc_write_block_metadata`, `zcashlc_free_block_meta, `zcashlc_free_blocks_meta`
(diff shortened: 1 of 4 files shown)


nuttycom (approved):
utACK

PR #89, opened 2023-04-26
[#83] Get Fingerprint for Seed bytes

### LINKED ISSUE #83: Get Fingerprint for Seed bytes
Implement functions that:

- receive seed bytes and return a seed fingerprint according to zip-32

https://zips.z.cash/zip-0032#seed-fingerprints

pacu:
Closes #83

This allow the SDK to use the seed phrase signature for identifiers that are coupled to it such as synchronizer aliases without revealing anything about the seed itself


nuttycom (approved):
utACK

PR #93, opened 2023-05-19
Rust updates

str4d:


rust/src/lib.rs
```diff
@@ -547,16 +576,15 @@ pub unsafe extern "C" fn zcashlc_init_blocks_table(
...
             hash,
             time,
             &sapling_tree,
-        ) {
-            Ok(()) => Ok(1),
-            Err(e) => Err(format_err!("Error while initializing blocks table: {}", e)),
-        }
+        )
+        .map(|()| 1)
+        .with_context(|| "Error while initializing blocks table")
```
pacu, line 587: This actually breaks the error we expect, because the database throws an "everything is fine" error here

rust/src/lib.rs
```diff
@@ -184,7 +217,7 @@ pub unsafe extern "C" fn zcashlc_init_data_database(
         match init_wallet_db(&mut db_data, seed) {
             Ok(_) => Ok(0),
             Err(MigratorError::Adapter(WalletMigrationError::SeedRequired)) => Ok(1),
-            Err(e) => Err(format_err!("Error while initializing data DB: {}", e)),
+            Err(e) => Err(e).with_context(|| "Error while initializing data DB"),
```
pacu, line 220: I think this context is being lost in the FFI. We only get "Error while initializing data DB" from the other side. We need to get similar output until we have rich types.

rust/src/lib.rs
```diff
@@ -282,7 +315,7 @@ pub unsafe extern "C" fn zcashlc_create_account(
                 let encoded = usk.to_bytes(Era::Orchard);
                 Box::into_raw(Box::new(FFIBinaryKey::new(account, encoded)))
             })
-            .map_err(|e| format_err!("Error while initializing accounts: {}", e))
+            .with_context(|| "Error while initializing accounts")
```
pacu, line 318: same as above

rust/src/lib.rs
```diff
@@ -393,12 +426,11 @@ pub unsafe extern "C" fn zcashlc_init_accounts_table_with_keys(
...
-            .map_err(|e| format_err!("Error decoding unified full viewing keys: {:?}", e))?;
+            .map_err(|e| anyhow!("Error decoding unified full viewing keys: {:?}", e))?;
 
-        match init_accounts_table(&db_data, &ufvks) {
-            Ok(()) => Ok(true),
-            Err(e) => Err(format_err!("Error while initializing accounts: {}", e)),
-        }
+        init_accounts_table(&db_data, &ufvks)
+            .map(|()| true)
+            .with_context(|| "Error while initializing accounts")
```
pacu, line 433: same as above

pacu (changes requested):
please check error handling specially around init functions to maintain same functionality or provide more context on how they should be handled now 🙏

pacu (changes requested):
please check error handling specially around init functions to maintain same functionality or provide more context on how they should be handled now 🙏

rust/src/lib.rs
str4d: Aha, the problem is that I was misled by [the crate's root-level documentation](https://docs.rs/anyhow/1.0.71/anyhow/#details) into believing that the default `Display` impl would show every level. There's [a different page in the documentation](https://docs.rs/anyhow/1.0.71/anyhow/struct.Error.html#display-representations) that states the default `Display` impl (`{}`) only prints the outer-most layer of the error, and we need to use the alternate instead (`{:#}`). However, that isn't accessible via the error's `ToString` implementation, and that is what the `ffi-helpers` crate we are using relies on.

Easy fix: I undo all the changes I thought were me being helpful and structuring the errors more correctly, and just use the `anyhow!()` macro everywhere we were using `format_err!()` before.

str4d: Put another way, it's not that the context is being lost in the FFI; it's that _only_ the context is being passed across the FFI.

str4d:
[Force-pushed](https://github.com/zcash-hackworks/zcash-light-client-ffi/compare/32cb823425a607b36a71a0bff1d72261fb46b030..6ec24acfc7d71c89b04f961cac2a4630be90b069) to remove the `with_context` usage, as it turned out to lose information when combined with the `ffi-helpers` crate.

rust/src/lib.rs
```diff
@@ -107,12 +113,44 @@ unsafe fn wallet_db(
...
+    log_panics::init();
+
+    // Manually build the Rayon thread pool, so we can name the threads.
+    rayon::ThreadPoolBuilder::new()
+        .thread_name(|i| format!("zc-rayon-{}", i))
+        .build_global()
+        .expect("Only initialized once");
+
+    debug!("Rust backend has been initialized successfully");
+    cfg_if::cfg_if! {
```
pacu, line 147: @str4d I think I'm not seeing any of these two messages being printed in the debugger console or in the OSLog console. They used to be .

rust/src/lib.rs
```diff
@@ -107,12 +113,44 @@ unsafe fn wallet_db(
...
+
+/// Initializes global Rust state, such as the logging infrastructure and threadpools.
+///
+/// # Panics
+///
+/// This method panics if called more than once.
+#[no_mangle]
+pub extern "C" fn zcashlc_init_on_load() {
+    // Set up the tracing layer for the Apple OS logging framework.
+    let os_layer = os_log::layer("co.electriccoin.ios", "rust").with_filter(LevelFilter::INFO);
```
pacu, line 132: Unfortunately I'm not seeing this reverse domain in any of the fields of the events reported by OS Logger.

pacu (changes requested):

rust/src/lib.rs
str4d: Oh, whoops: I'm setting the max log level to INFO above, so DEBUG level log messages would be completely ignored. I'll fix the PR to not limit log levels at all on the Rust side, and hopefully the Apple side limiting will work.

pacu (approved):

str4d:
@pacu tested this and saw the following tree of categories:

image

I interpret this as:
- `rust` from the custom logger's name.
  - `rust` as the hard-coded name for all Signposts (because the Apple SDK forces it to be a literal).
    - Span name. We aren't getting any nesting of spans, so you need to just know how the spans are nested (e.g. that `scan_cached_blocks` time covers all of the other three combined).
      - Span name again? Maybe this is because no `os_signpost` events are being generated, so it inserts a synthetic one corresponding to the interval. Or maybe these correspond to the beginning and end of the interval, IDK.

PR #94, opened 2023-07-19
Migrate to Rust crate versions with fast spendability support

### LINKED ISSUE #95: Migrate to Rust crate versions with fast spendability support

str4d:
Closes #95.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,59 @@
-# 0.3.1
+# Changelog
+All notable changes to this library will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## 0.4.0-rc.1 - 2023-09-08
+
+### Notable Changes
+
+This release overhauls the FFI library to provide support for allowing wallets to
+spend funds without fully syncing the blockchain. This results in significant 
+changes to much of the API; it is recommended that users review the changes
+from the previous release carefully.
+
+### Changed
+- `anyhow` is now used for error management
+
+### Issues Resolved
+- [#95] Update to `zcash_client_backend` and `zcash_client_sqlite` with fast sync support
+
+Updated dependencies:
+  - `zcash_address 0.3`
+  - `zcash_client_backend 0.10.0-rc.1`
+  - `zcash_client_sqlite 0.8.0-rc.1`
+  - `zcash_primitives 0.13.0-rc.1`
+  - `zcash_proofs 0.13.0-rc.1`
+
+  - `orchard 0.6`
+  - `ffi_helpers 0.3`
+  - `secp256k1 0.26`
+
+Added dependencies:
+  - `anyhow 0.1`
+  - `prost 0.12`
+  - `cfg-if 1.0`
+  - `rayon 1.7`
+  - `log-panics 2.0`
+  - `once_cell 1.0`
+  - `sharded-slab 0.1`
+  - `tracing 0.1`
+  - `tracing-subscriber 0.3`
+
+## 0.3.1
 - [#88] unmined transaction shows note value spent instead of tx value
 
 Fixes an issue where a sent transaction would show the whole note spent value
 instead of the value of that the user meant to transfer until it was mined.
 
-# 0.3.0
+## 0.3.0
+
 - [#87] Outbound transactions show the wrong amount on v_transactions
 
-removes `v_tx_received` and `v_tx_sent`. 
+removes `v_tx_received` and `v_tx_sent`.
 
-`v_transactions` now shows the `account_balance_delta` column where the clients can 
+`v_transactions` now shows the `account_balance_delta` column where the clients can
 query the effect of a given transaction in the account balance. If fee was paid from
 the account that's being queried, the delta will include it. Transactions where funds
 are received into the queried account, will show the amount that the acount is receiving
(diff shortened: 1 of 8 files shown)


rust/src/lib.rs
```diff
@@ -1864,13 +1803,252 @@ pub unsafe extern "C" fn zcashlc_rewind_to_height(
...
+/// # Safety
+///
+/// - `root_hash_ptr` must be non-null and must be valid for reads for `root_hash_ptr_len`
+///   bytes, and it must have an alignment of `1`.
+/// - The total size `root_hash_ptr_len` of the slice pointed to by `root_hash_ptr` must
+///   be no larger than `isize::MAX`. See the safety documentation of `pointer::offset`.
+#[repr(C)]
+pub struct FfiSubtreeRoot {
+    root_hash_ptr: *mut u8,
+    root_hash_ptr_len: usize,
```
nuttycom, line 1824: Why a pointer and a length here? These should all be pointers to 32 byte arrays, right?

rust/src/lib.rs
```diff
@@ -1864,13 +1803,252 @@ pub unsafe extern "C" fn zcashlc_rewind_to_height(
...
+            .map(|()| true)
+            .map_err(|e| anyhow!("Error while storing Sapling subtree roots: {}", e))
+    });
+    unwrap_exc_or(res, false)
+}
+
+/// Rewinds the data database to the given height.
+///
+/// If the requested height is greater than or equal to the height of the last scanned
+/// block, this function does nothing.
```
nuttycom, line 1905: This comment seems wrong. Copy-paste error?

nuttycom (changes requested):
Overall this looks good; changes requested for one documentation error plus a question about a FFI representation.

rust/src/lib.rs
str4d: I copied this FFI pattern from `FFIBlockMeta` where it is used for block hashes. But it is indeed currently the case that all root hashes we'd see here are 32 bytes. I'll leave this one as-is since it's been tested and changing it would require more changes to the Kit, but I'll add subsequent similar fields without the length.

rust/src/lib.rs
```diff
@@ -1864,13 +1712,255 @@ pub unsafe extern "C" fn zcashlc_rewind_to_height(
...
+            .collect::<Vec<_>>();
+
+        Ok(FfiScanRanges::ptr_from_vec(ffi_ranges))
+    });
+    unwrap_exc_or_null(res)
+}
+
 /// Scans new blocks added to the cache for any transactions received by the tracked
-/// accounts.
+/// accounts, while checking that they form a valid chan.
```
pacu, line 1956:
```suggestion
/// accounts, while checking that they form a valid chain.
```

rust/src/lib.rs
```diff
@@ -1864,13 +1712,255 @@ pub unsafe extern "C" fn zcashlc_rewind_to_height(
...
+/// A struct that contains a subtree root.
+///
+/// # Safety
+///
+/// - `root_hash_ptr` must be non-null and must be valid for reads for `root_hash_ptr_len`
+///   bytes, and it must have an alignment of `1`.
+/// - The total size `root_hash_ptr_len` of the slice pointed to by `root_hash_ptr` must
+///   be no larger than `isize::MAX`. See the safety documentation of `pointer::offset`.
+#[repr(C)]
+pub struct FfiSubtreeRoot {
```
pacu, line 1724: suggestion: check that other structs have the FFI all in caps.

str4d: I followed Rust making convention here, where acronyms are CamelCased (`HtmlFoo` instead of `HTMLFoo`).

pacu: I assumed so! It's kind of a philosophical questions, should the FFI target swift generator's conventions or Rust's? 

I don't have a preference, I'd just pick one or the other.

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
```diff
@@ -52,6 +52,89 @@ typedef struct FFIEncodedKeys {
...
+  struct FfiSubtreeRoot *ptr;
+  uintptr_t len;
+} FfiSubtreeRoots;
+
+/**
+ * A struct that contains details about scan progress.
+ *
+ * When `denominator` is zero, the numerator encodes a non-progress indicator:
+ * - 0: progress is unknown.
+ * - 1: an error occurred.
```
daira, line 99: Another way to do this would be to call the struct FfiScanProgressResult, and have another field that contains a status enum (None, Some, or Error).

nuttycom (approved):
utACK https://github.com/zcash-hackworks/zcash-light-client-ffi/pull/94/commits/0fcb534bf53c27c7b2e9a38e61a01beee2eabe31

CHANGELOG.md
```diff
@@ -1,34 +1,79 @@
-# 0.3.1
+# Changelog
+All notable changes to this library will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## 0.4.0-rc.1 - 2023-09-08
```
str4d, line 7: Version in `rust/Cargo.toml` was not updated before the `make xcframework`.

rust/src/lib.rs
```diff
@@ -308,16 +323,34 @@ pub unsafe extern "C" fn zcashlc_create_account(
...
+        let recover_until = recover_until.try_into().ok();
+
+        let birthday =
+            AccountBirthday::from_treestate(treestate, recover_until).map_err(|e| match e {
+                BirthdayError::HeightInvalid(e) => {
+                    anyhow!("Invalid TreeState: Invalid height: {}", e)
+                }
+                BirthdayError::Decode(e) => {
+                    anyhow!("Invalid TreeState: Invalid frontier encoding: {}", e)
+                }
```
str4d, line 349: Non-blocking: this should have been part of a `Display` impl on `BirthdayError` (I worked around it here).

str4d (commented):
LGTM. Needs version bump.

str4d (commented):
ACK

PR #98, opened 2023-09-12
Fix checkpoint depth bug

nuttycom:
See [zcash/librustzcash#963]

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -893,7 +893,7 @@ dependencies = [
 
 [[package]]
 name = "libzcashlc"
-version = "0.4.0-rc.1"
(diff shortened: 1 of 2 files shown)


daira (approved):
I can't review the binary files, but I checked that the librustzcash commit (f2dcba3a34bb5f00ec670f478029621201529b2d) is the one that I reviewed for zcash/librustzcash#963 .

daira:
Ideally 'make xcframework' would be automatic. Having it be done manually by the committer is error-prone and results in PRs that can't be fully reviewed, which is a security hazard.

str4d (approved):
utACK 7c4e57d8225168bfe4e004ff9da1206d75a2fea0

PR #105, opened 2023-11-08
Implement transfer proposal functions

tw0po1nt:
This is a very wip first shot at implementing the proposal mechanism in the FFI layer.

Known items that will change:

- Missing documentation for new API. I held off in case of major PR review changes. TODO as a PR feedback item.
- Crate patch will need to go away and the proper crate refs updated, once [this PR](https://github.com/zcash/librustzcash/pull/891) is merged.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -56,3 +56,10 @@ lto = true
 [features]
 mainnet = ["zcash_client_sqlite/mainnet"]
 testnet = []
+
+[patch.crates-io]
+zcash_address = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
+zcash_client_backend = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
+zcash_client_sqlite = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
+zcash_primitives = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
+zcash_proofs = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -58,8 +58,8 @@ mainnet = ["zcash_client_sqlite/mainnet"]
 testnet = []
 
 [patch.crates-io]
-zcash_address = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
-zcash_client_backend = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
-zcash_client_sqlite = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
-zcash_primitives = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
-zcash_proofs = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "236cd569ee4a824d98920d1a61b8cc2b90ceba1d" }
+zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "236cd569ee4a824d98920d1a61b8cc2b90ceba1d" }
+zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "236cd569ee4a824d98920d1a61b8cc2b90ceba1d" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "236cd569ee4a824d98920d1a61b8cc2b90ceba1d" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "236cd569ee4a824d98920d1a61b8cc2b90ceba1d" }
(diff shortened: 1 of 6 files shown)


rust/Cargo.toml
```diff
@@ -56,3 +56,10 @@ lto = true
 [features]
 mainnet = ["zcash_client_sqlite/mainnet"]
 testnet = []
+
+[patch.crates-io]
+zcash_address = { git = "https://github.com/nuttycom/librustzcash.git", rev = "8be0d8487495dd8dd4c1a54177b7418547f30060" }
```
tw0po1nt, line 61: Obviously this will need to be removed once this revision is properly merge into librustzcash.

tw0po1nt (commented):
Brief self-review completed

rust/src/lib.rs
```diff
@@ -2679,6 +2784,107 @@ pub unsafe extern "C" fn zcashlc_shield_funds(
...
+    memo: *const u8,
+    shielding_threshold: u64,
+    network_id: u32,
+    min_confirmations: u32,
+    use_zip317_fees: bool,
+    proposal_bytes_ret: *mut u8
+) -> bool {
+    let res = catch_panic(|| {
+        let network = parse_network(network_id)?;
+        let min_confirmations = NonZeroU32::new(min_confirmations)
```
nuttycom, line 2802: The upstream APIs have changed to allow zero-conf shielding transactions to be created, so we should just make this a `u32`.

rust/src/lib.rs
```diff
@@ -2679,6 +2784,107 @@ pub unsafe extern "C" fn zcashlc_shield_funds(
...
+
+        let shielding_threshold = NonNegativeAmount::from_u64(shielding_threshold)
+            .map_err(|()| anyhow!("Invalid amount, out of range"))?;
+
+        let account = db_data
+            .get_account_for_ufvk(&usk.to_unified_full_viewing_key())?
+            .ok_or_else(|| anyhow!("Spending key not recognized."))?;
+
+        let taddrs: Vec<TransparentAddress> = db_data
+            .get_target_and_anchor_heights(NonZeroU32::MIN)
```
nuttycom, line 2823: This isn't really an anchor height; you should probably just use `chain_tip - min_confirmations`.

nuttycom (commented):
Overall this looks great. [zcash/librustzcash#891] should merge this morning, and at that point this can be updated to point at the current `main` commit and fix any issues raised by API changes.

rust/src/lib.rs
tw0po1nt: I'm confused about this one. This is just what zcashls_shield_funds was doing. Are you saying instead of making the call to `db_data. get_target_and_anchor_heights` to just pass `chain_tip - min_confirmations` to db_data.get_transparent_balances()`?

nuttycom: Yup! You'll notice that that implementation has changed: https://github.com/zcash/librustzcash/blob/236cd569ee4a824d98920d1a61b8cc2b90ceba1d/zcash_client_backend/src/data_api/wallet.rs#L505-L519

nuttycom: You can also just use the `propose_shielding` function directly.

rust/src/lib.rs
```diff
@@ -2509,6 +2498,172 @@ pub unsafe extern "C" fn zcashlc_create_to_address(
...
+        }));
+
+        let prover = LocalTxProver::new(spend_params, output_params);
+
+        let txid = create_proposed_transaction(
+            &mut db_data,
+            &network,
+            prover,
+            &usk,
+            OvkPolicy::Sender,
```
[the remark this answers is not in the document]
tw0po1nt: Updated, let me know what you think of this approach.

rust/src/lib.rs
```diff
@@ -2509,6 +2498,172 @@ pub unsafe extern "C" fn zcashlc_create_to_address(
...
+
+        propose_transfer(
+            &mut db_data,
+            &network,
+            account,
+            &input_selector,
+            req,
+            min_confirmations,
+        )
+        .map(|proposal| {
```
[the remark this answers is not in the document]
tw0po1nt: I think the refactor I did to simply return a bool and use a `*mut u8` ptr param has resolved this?

rust/src/lib.rs
tw0po1nt: Updated.

rust/src/lib.rs
```diff
@@ -2679,6 +2792,105 @@ pub unsafe extern "C" fn zcashlc_shield_funds(
...
+            >| anyhow!("Error while proposing transfer: {}", e),
+        )?;
+
+        if let Some(proto) = proposal::Proposal::from_standard_proposal(&network, &proposal) {
+            let mut buf = Vec::new();
+            buf.reserve(proto.encoded_len());
+            proto.encode(&mut buf).unwrap();
+            let proposal_bytes = Box::new(buf);
+            unsafe { std::ptr::copy(proposal_bytes.as_ptr(), proposal_bytes_ret, proposal_bytes.len()); }
+            Ok(true)
```
str4d, line 2886: This would almost certainly have caused a crash. In this code, `proposal_bytes_ret` is allocated by the Swift caller, but transaction proposal encodings are variable length (and protobuf encodings are non-canonical so can even vary in length with re-encoding), and there is no way for the caller to know what length they need to allocate.

In #111 I address this by allocating the memory on the Rust side.

str4d:
Replaced by #110 and #111.

rust/src/lib.rs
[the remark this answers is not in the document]
str4d: This is not addressed by existing PRs; it would be useful to extract these changes out into a separate PR.

rust/src/lib.rs
```diff
@@ -2509,6 +2498,172 @@ pub unsafe extern "C" fn zcashlc_create_to_address(
...
+pub unsafe extern "C" fn zcashlc_create_proposed_transfer(
+    db_data: *const u8,
+    db_data_len: usize,
+    usk_ptr: *const u8,
+    usk_len: usize,
+    proposal: *const FFIRawProtobuf,
+    spend_params: *const u8,
+    spend_params_len: usize,
+    output_params: *const u8,
+    output_params_len: usize,
```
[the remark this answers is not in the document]
str4d: Yes: app install size is a significant concern for both app stores and users (source: discussions with mobile developers back in 2019/2020), so most applications with large artifacts (like games) keep their base app size small and then separately download any necessary artifacts.

PR #106, opened 2023-12-19
Rust changes for 0.5.0

str4d:


str4d:
In a pairing with @nuttycom, we decided to try a feature branch-ish approach for 0.5.0:

- I will create two branches:
  - A feature branch based off `main`.
  - A "release preview" branch based off the feature branch.
- We create PRs that update the Rust code and the FFI APIs, targeting this feature branch. These PRs _do not_ contain `make xcframework` commits.
- Periodically when necessary, we merge the feature branch into the release preview branch, and then add a `make xcframework` commit there.
- Over in the Swift SDK repo, we have a similar feature branch. It refers to commits in the release preview branch (so it has working FFI libraries). Whenever we bump the revision, we also bump a comment in the package file saying which revision of the feature branch this release preview commit was built from.
- When it comes time to cut the release:
  - We push a `make xcframework` commit directly to the feature branch.
  - We **delete the release preview branch**.
  - We merge the feature branch to `main` and tag the release.
  - In the Swift SDK repo, we replace whatever revision it currently uses with the tagged release, and then merge its feature branch.

What this means is that, for tagged releases of the Swift SDK, it will only depend on revisions of this repository containing binary data for tagged releases of the FFI backend (and in particular, only binary data for things we support). This means that checking out commits from the middle of the feature branch (e.g. during a bisect) will become broken (as the commits they point to will have been deleted and GCed from this repo), but the code comment in the Swift SDK repo provides a pathway to locally recreating the SDK build (check out the commit from the code comment, run `make xcframework` locally, and then alter the Swift SDK to use the FFI repo at the local path instead of via URL).

str4d:
Replaced by #112.

PR #103, opened 2023-10-19
Update dependencies

ebfull:

PR #107, opened 2024-01-10
Merge 0.4.1 release branch

str4d:


str4d:
I was initially worried that the release PR did not run a `make xcframework`, but then realised that I had done so in #103, so the changes in #104 were just to bump metadata that is not visible in the binaries.

PR #109, opened 2024-01-10
Conditionally compile Apple-specific APIs

str4d:
This enables development on non-Apple machines, and potentially in future usage of the Swift SDK with non-Apple targets.


nuttycom (approved):
utACK 04106049ee256d576283753c99ea32004ce1ba33

PR #112, opened 2024-01-12
Feature branch for 0.5.0

str4d:


rust/src/lib.rs
```diff
@@ -2353,14 +2406,73 @@ pub unsafe extern "C" fn zcashlc_decrypt_and_store_transaction(
...
-/// Do not call this multiple times in parallel, or you will generate transactions that
-/// double-spend the same notes.
+/// - `ptr` must be non-null and valid for reads for `len` bytes, and it must have an
+///   alignment of `1`. Its contents must be an encoded Proposal protobuf.
+/// - The memory referenced by `ptr` must not be mutated for the lifetime of the struct
+///   (up until [`zcashlc_free_boxed_slice`] is called with it).
+/// - The total size `len` must be no larger than `isize::MAX`. See the safety
+///   documentation of `pointer::offset`.
+#[repr(C)]
+pub struct FfiBoxedSlice {
```
pacu, line 2437: If "Its contents must be an encoded Proposal protobuf" shouldn't this have a meaningful name indicating this so that it's not used for other purposes?

pacu (changes requested):

rust/src/lib.rs
str4d: I think I added that line to the documentation before I decided to implement this as an `FfiBoxedSlice`. I think it's better to just remove that requirement and allow this to contain an arbitrary byte slice, the contents of which is determined by the method it is returned from.

rust/src/lib.rs
```diff
@@ -1834,7 +1790,94 @@ pub struct FfiScanProgress {
...
+/// - `scan_progress` must, if non-null, point to a struct having the layout of
+///   [`FfiScanProgress`].
+#[repr(C)]
+pub struct FfiWalletSummary {
+    account_balances: *mut FfiAccountBalance,
+    account_balances_len: usize,
+    chain_tip_height: i32,
+    fully_scanned_height: i32,
+    scan_progress: *mut FfiScanProgress,
+    next_sapling_subtree_index: u64,
```
str4d, line 1824: The protobuf constrains this to a `u32`, so we should do that here as well. I'll fix this in the next PR.

str4d: Opened #118.

str4d (commented):
ACK 9a987c24240992a5eb3ff447e93f581d49a23acd.

PR #115, opened 2024-01-24
[#114] Create a transaction proposal from ZIP-321 payment URI

### LINKED ISSUE #114: Create a transaction proposal from ZIP-321 payment URI

pacu:
closes #114

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -2564,6 +2564,65 @@ pub unsafe extern "C" fn zcashlc_propose_transfer(
     unwrap_exc_or_null(res)
 }
 
+/// Select transaction inputs, compute fees, and construct a proposal for a transaction
+/// from a ZIP-321 payment URI that can then be authorized and made ready for submission to the
+/// network with `zcashlc_create_proposed_transaction`.
+///
+/// # Safety
+///
+/// - `db_data` must be non-null and valid for reads for `db_data_len` bytes, and it must have an
+///   alignment of `1`. Its contents must be a string representing a valid system path in the
+///   operating system's preferred representation.
+/// - The memory referenced by `db_data` must not be mutated for the duration of the function call.
+/// - The total size `db_data_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of pointer::offset.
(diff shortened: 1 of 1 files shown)


rust/src/lib.rs
```diff
@@ -2564,6 +2564,65 @@ pub unsafe extern "C" fn zcashlc_propose_transfer(
...
+/// - `use_zip317_fees` `true`` to use ZIP-317 fees.
+#[no_mangle]
+pub unsafe extern "C" fn zcashlc_propose_transfer_from_uri(
+    db_data: *const u8,
+    db_data_len: usize,
+    account: i32,
+    payment_uri: *const c_char,
+    network_id: u32,
+    min_confirmations: u32,
+    use_zip317_fees: bool,
```
nuttycom, line 2591: Does it make sense to have a `use_zip317_fees` argument? Or should we always use ZIP 317 fees for this API?

pacu: I'd say we remove it! Would you like me to apply the same change to the rest of the APIs that have this boolean flag?

rust/src/lib.rs
```diff
@@ -2564,6 +2564,65 @@ pub unsafe extern "C" fn zcashlc_propose_transfer(
...
+) -> *mut FfiBoxedSlice {
+    let res = catch_panic(|| {
+        let network = parse_network(network_id)?;
+        let min_confirmations = NonZeroU32::new(min_confirmations)
+            .ok_or(anyhow!("min_confirmations should be non-zero"))?;
```
nuttycom, line 2596: I suspect that this is also a problem elsewhere, but I'm just noticing that `min_confirmations` should actually be bounded to the range `1..=100` because we only routinely retain 100 blocks worth of checkpoints that we can construct anchors from.

nuttycom (approved):
utACK 4cc4347aee4785173a21827164a288aaa9ffe7dc

rust/src/lib.rs
str4d: Not yet.

str4d (approved):
utACK 4cc4347aee4785173a21827164a288aaa9ffe7dc

PR #121, opened 2024-03-06
Feature branch for 0.6.0

str4d:


rust/src/lib.rs
```diff
@@ -2743,9 +2783,39 @@ pub unsafe extern "C" fn zcashlc_propose_shielding(
                         )
                     })
             })?
-            .keys()
-            .cloned()
-            .collect();
+            .into_iter();
+
+        let from_addrs = if let Some((receiver_if_none_provided, value)) = account_receivers.next()
```
nuttycom, line 2788: This is a weird way to approach this, to me. Shouldn't we be checking that the map returned by `get_transparent_balances` has length 1 and contains the transparent receiver that is provided first, and otherwise just taking the first? I don't understand the rationale for approaching the address selection in this fashion.

rust/src/lib.rs
```diff
@@ -2745,9 +2783,39 @@ pub unsafe extern "C" fn zcashlc_propose_shielding(
...
+                }
+            } else {
+                return Err(anyhow!(
+                        "Account has more than one transparent receiver with funds to shield; this is not yet supported by the SDK. Provide a specific transparent receiver to shield funds from."
+                    ));
+            }
+        } else {
+            // There are no transparent funds to shield; don't create a proposal.
+            return Ok(FfiBoxedSlice::none());
+        };
```
str4d, line 2818: I've tried rewriting this to use the `HashMap` directly. It is almost as complex, and now has an unwrap.
```suggestion
        let from_addrs = match account_receivers.len() {
            // There are no transparent funds to shield; don't create a proposal.
            0 => return Ok(FfiBoxedSlice::none()),
            // Only one transparent receiver has funds to shield.
            1 => {
                if let Some(addr) = transparent_receiver {
                    match account_receivers.get(&addr) {
                        // The provided transparent receiver has sufficient funds to shield.
                        Some(&value) if value >= shielding_threshold.into() => [addr],
                        // No or insufficient funds to shield; don't create a proposal.
                        None | Some(_) => return Ok(FfiBoxedSlice::none()),
                    }
                } else {
                    let (addr, value) = account_receivers.into_iter().next().expect("has entry");
                    if value >= shielding_threshold.into() {
                        // The only transparent receiver has sufficient funds to shield.
                        [addr]
                    } else {
                        // Insufficient funds to shield; don't create a proposal.
                        return Ok(FfiBoxedSlice::none());
                    }
                }
            }
            _ => {
                return Err(format_err!(
                    "Account has more than one transparent receiver with funds to shield; this is not yet supported by the SDK. Provide a specific transparent receiver to shield funds from."
                ));
            }
        };
```

nuttycom: What about this: 
```rust
         let from_addrs = if account_receivers.len() > 1 {
            return Err(anyhow!(
                    "Account has more than one transparent receiver with funds to shield; this is not yet supported by the SDK. Provide a specific transparent receiver to shield funds from."
                ));
        } else {
            if let Some((addr, value)) = transparent_reciver
                .and_then(|addr| account_receivers.get(addr).map(|value| (addr, value)))
                .or_else(|| account_receivers.iter().next())
            {
                if value >= shielding_threshold.into() {
                    // ... and it has sufficient funds to shield.
                    [*addr]
                } else {
                    // Insufficient funds to shield; don't create a proposal.
                    return Ok(FfiBoxedSlice::none());
                }
            } else {
                // There are no transparent funds to shield; don't create a proposal.
                return Ok(FfiBoxedSlice::none());
            }
        };
```

str4d: I've now gone back to my iterator approach but leveraging the length-matching approach I used above for the `HashMap`, and it's much cleaner:
```suggestion
        let from_addrs = match (
            transparent_receiver,
            account_receivers.next(),
            account_receivers.next(),
        ) {
            // There are no transparent funds to shield.
            (_, None, None) => return Ok(FfiBoxedSlice::none()),
            // There is only one transparent receiver and it has insufficient funds to
            // shield. It is irrelevant whether this is the provided receiver; if it is
            // not, then the provided receiver has no funds to shield.
            (Some(_), Some((_, value)), None) if value < shielding_threshold.into() => {
                return Ok(FfiBoxedSlice::none())
            }
            // The provided transparent receiver has no funds to shield.
            (Some(addr), Some((addr_with_funds, _)), None) if addr != addr_with_funds => {
                return Ok(FfiBoxedSlice::none())
            }
            // The provided transparent receiver has sufficient funds to shield.
            (Some(addr), Some(_), None) => [addr],
            // The only transparent receiver has sufficient funds to shield.
            (None, Some((addr_with_funds, _)), None) => [addr_with_funds],
            _ => {
                return Err(format_err!(
                    "Account has more than one transparent receiver with funds to shield; this is not yet supported by the SDK. Provide a specific transparent receiver to shield funds from."
                ));
            }
        };
```

nuttycom: https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/122

nuttycom (approved):
utACK 7c801be1f445402a433b32835a50d832e8a50437

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
```diff
@@ -259,6 +259,28 @@ typedef struct FfiBoxedSlice {
...
 } FfiBoxedSlice;
 
+/**
+ * A struct that contains a pointer to, and length information for, a heap-allocated
+ * slice of `[u8; 32]` arrays.
+ *
+ * # Safety
+ *
+ * - `ptr` must be non-null and must be valid for reads for `len * mem::size_of::<[u8; 32]>()`
+ *   many bytes, and it must be properly aligned. This means in particular:
```
daira, line 269: As I read [the relevant Rust documentation](https://doc.rust-lang.org/reference/type-layout.html#array-layout), the alignment of `[u8; 32]`, or a slice of it, is guaranteed to be 1 (and, by applying the "`[T; n]` has the same alignment as `T`" rule twice, this is also true of `[[u8; 32]; len]` or a slice of it). Therefore "must be properly aligned" is vacuous here.

rust/src/lib.rs
```diff
@@ -2493,6 +2493,13 @@ impl FfiBoxedSlice {
         let slim_ptr: *mut u8 = fat_ptr as _;
         Box::into_raw(Box::new(FfiBoxedSlice { ptr: slim_ptr, len }))
     }
+
+    fn none() -> *mut Self {
+        Box::into_raw(Box::new(Self {
+            ptr: ptr::null_mut(),
+            len: 0,
+        }))
+    }
```
daira, line 2502: I meant to disagree with changing `FfiBoxedSlice` to allow wrapping a null pointer, but I don't think I got around to submitting that comment :-/ If anything I think we should do the opposite: explicitly disallow null pointers. We don't need to use `Ok(FfiBoxedSlice::none())`; we can use an error instead.

daira (commented):
Post-hoc ACK modulo comments. I think we need to discuss https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/121/files#r1517147614

PR #123, opened 2024-03-14
Feature branch for 0.7.0

str4d:
- [x] Migrated to new accounts backend.
- [x] Seed relevance API.
- [x] Orchard enabled.
- ~~Payments to TEX addresses supported.~~


rust/src/lib.rs
```diff
@@ -125,13 +124,46 @@ fn block_db(fsblock_db: *const u8, fsblock_db_len: usize) -> anyhow::Result<FsBl
...
+    db_data: &WalletDb<rusqlite::Connection, P>,
+    account_index: i32,
+) -> anyhow::Result<AccountId> {
+    let requested_account_index = account_id_from_i32(account_index)?;
+
+    // Find the single account matching the given ZIP 32 account index.
+    let mut accounts = db_data
+        .get_account_ids()?
+        .into_iter()
+        .filter_map(|account_id| {
```
nuttycom, line 144: Wouldn't `find_map` be better here?

rust/src/lib.rs
```diff
@@ -355,17 +387,58 @@ pub unsafe extern "C" fn zcashlc_create_account(
...
+) -> bool {
+    let res = catch_panic(|| {
+        let network = parse_network(network_id)?;
+        let db_data = unsafe { wallet_db(db_data, db_data_len, network)? };
+        let seed = Secret::new((unsafe { slice::from_raw_parts(seed, seed_len) }).to_vec());
+
+        for account_id in db_data.get_account_ids()? {
+            if !db_data.validate_seed(account_id, &seed)? {
+                return Ok(false);
+            }
```
nuttycom, line 434:
```suggestion
            if db_data.validate_seed(account_id, &seed)? {
                return Ok(true);
            }
```
Bug: this method will currently return `true` if the database contains no accounts.

rust/src/lib.rs
```diff
@@ -355,17 +387,58 @@ pub unsafe extern "C" fn zcashlc_create_account(
...
+        }
+
+        Ok(true)
```
nuttycom, line 437:
```suggestion
        Ok(false)
```

rust/src/lib.rs
```diff
@@ -1828,12 +1901,29 @@ pub struct FfiWalletSummary {
...
+    ) -> *mut Self {
         let (account_balances, account_balances_len) = {
             let account_balances: Box<[FfiAccountBalance]> = summary
                 .account_balances()
                 .iter()
-                .map(FfiAccountBalance::new)
+                .map(|(account_id, balance)| {
+                    let account_index = match db_data
+                        .get_account(*account_id)
+                        .expect("we got the summary from the wallet")
```
nuttycom, line 1915: This expectation should probably be an error; it indicates database corruption or failure of a query that is separate from the summary retrieval.

nuttycom (changes requested):
One blocking bug.

rust/src/lib.rs
str4d: No, I want it to be an error if there is more than one matching ZIP 32 account index.

rust/src/lib.rs
str4d: Also, your suggestion has a bug: it early-returns, so it only checks that the _first_ account is valid with the seed, and not that _all_ accounts are valid with the seed.

rust/src/lib.rs
str4d: Ah yep, this specific exception could be a transient error. The `Option::unwrap` can still be used.

rust/src/lib.rs
nuttycom: That's the intention; it's future-proof (because for imported UFVKs we won't want to validate against them.)

str4d: Okay, I think this indicates that we do not have the same idea about what this API is meant to validate.

My understanding is that at the mobile SDK level, we currently only support a single seed phrase, and only derived accounts. Therefore, `validateSeed` checks that the database only contains accounts corresponding to this single seed phrase. If it contains _any_ accounts for a different seed phrase, or an unknown seed phrase, then we cannot use the database. _If_ we move to supporting imported UFVKs, then at that time we'd weaken this to ignore imported accounts, but we would still check all derived accounts (and from that perspective, the `WalletRead::validate_seed` API is insufficient: it combines "account is imported" and "account is derived but not from this seed" as a single `Ok(false)`, whereas they need to be distinguishable in this case).

It seems that you instead want an API that checks if the given seed phrase is _relevant_ to the database, i.e. if _any_ account can be used with that seed phrase. That might be a separately useful API, but it doesn't solve the problem of "can I use this database backup", because if the caller does not know every seed phrase for every derived account, then the answer is no.

nuttycom: > It seems that you instead want an API that checks if the given seed phrase is relevant to the database, i.e. if any account can be used with that seed phrase. That might be a separately useful API, but it doesn't solve the problem of "can I use this database backup", because if the caller does not know every seed phrase for every derived account, then the answer is no.

That's not necessary for the seed phrase to be useful. Remember that when we go to spend, we take the USK as input, derive the UFVK, and then use _that_ to determine what funds we have available to spend. So we can have a database that contains UFVKs that don't correspond to the seed, and that's just fine; we won't be able to spend from those accounts.

rust/src/lib.rs
nuttycom: Ah, okay, I glossed over that `accounts.next()` was being called twice.

nuttycom (approved):
utACK 7860e5e88d3c10245ef3e5030f5fb94277173cb5

nuttycom (approved):
utACK 70cce1272c26ed52fbe

rust/src/lib.rs
```diff
@@ -355,17 +500,68 @@ pub unsafe extern "C" fn zcashlc_create_account(
...
     });
     unwrap_exc_or_null(res)
 }
 
+/// Checks whether the given seed is relevant to any of the accounts in the wallet.
+///
+/// Returns:
+/// - `1` for `Ok(true)`.
+/// - `0` for `Ok(false)`.
+/// - `-1` for `Err(_)`.
```
daira, line 527: Add safety documentation of arguments.

rust/src/lib.rs
```diff
@@ -355,17 +500,68 @@ pub unsafe extern "C" fn zcashlc_create_account(
...
+    let res = catch_panic(|| {
+        let network = parse_network(network_id)?;
+        let db_data = unsafe { wallet_db(db_data, db_data_len, network)? };
+        let seed = Secret::new((unsafe { slice::from_raw_parts(seed, seed_len) }).to_vec());
+
+        for account_id in db_data.get_account_ids()? {
+            match db_data.validate_seed(account_id, &seed) {
+                // The seed is relevant to this account. No need to check any others.
+                Ok(true) => return Ok(1),
+                // We know by dead reckoning that this account ID exists in the wallet, so
```
daira, line 545: Not sure what "by dead reckoning" means here.
```suggestion
                // We're looping over account IDs that we know exist in the wallet, so
```

daira (approved):
utACK

str4d (commented):
utACK c918662aeb23b429abb400558483f9927dc56245

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
```diff
@@ -396,6 +462,20 @@ struct FFIBinaryKey *zcashlc_create_account(const uint8_t *db_data,
                                             int64_t recover_until,
                                             uint32_t network_id);
 
+/**
+ * Checks whether the given seed is relevant to any of the accounts in the wallet.
+ *
+ * Returns:
+ * - `1` for `Ok(true)`.
+ * - `0` for `Ok(false)`.
```
daira, line 470: Non-blocking: Ideally, documentation at this level would not depend on knowing the semantics of the corresponding Rust function (which isn't linked here).

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
```diff
@@ -1013,6 +1093,8 @@ struct FfiScanSummary *zcashlc_scan_blocks(const uint8_t *fs_block_cache_root,
                                            const uint8_t *db_data,
                                            uintptr_t db_data_len,
                                            int32_t from_height,
+                                           const uint8_t *from_state,
+                                           uintptr_t from_state_len,
```
daira, line 1097: These new parameters have not been documented in the safety section. Non-blocking but please file an issue if it isn't fixed now.

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/libzcashlc
daira, line 1: Binary file not reviewed.

releases/XCFramework/libzcashlc.xcframework/ios-arm64_x86_64-simulator/libzcashlc.framework/libzcashlc
daira, line 1: Binary file not reviewed.

daira (approved):
utACK. Comments are non-blocking, but please file an issue for the safety documentation comments if you merge as-is. [Done in #124.]

PR #125, opened 2024-03-26
zcash_light_client_ffi release version 0.7.1

nuttycom:
This fixes an error in version 0.7.0 that could cause problems for wallets sending notes to their own external addresses.


str4d (approved):
utACK c7e5158edf5e62af15492d30237163b78af35ce9

daira (commented):
Post-hoc ACK

PR #130, opened 2024-03-29
Send change to Orchard by default

nuttycom:
This also modifies shielding behavior so that shielded funds are sent to the Orchard pool instead of to Sapling.


str4d (approved):
ACK db890e8bc52b04d26110de7f9a1ebdbe4d35741c

daira (commented):
Post-hoc ACK. It would be nice to have a test for the default, though.

PR #131, opened 2024-03-29
FFI 0.7.5

str4d:

PR #133, opened 2024-04-15
Set `MinimumOSVersion` to a wildly-high value

str4d:
For whatever reason, this works around a bug in XCode 15.3.

Part of Electric-Coin-Company/zashi-ios#1166.


nuttycom (approved):
tested ACK 27b6eb2aebfb06d22ccc98c0c03f88c730f10472

str4d:
Un-merged from #131 and re-opened as #135.

PR #135, opened 2024-04-16
Add dummy `Info.plist` to platform frameworks

str4d:
A bug in the XCode 15.3 build system causes a dummy framework folder to be included in the app bundle when it depends on a statically linked framework (like this one). This causes app validation to fail both in the simulator and when pushing an app build to the App Store Connect.

We work around the bug by adding a dummy `Info.plist` to the platform frameworks, with a `MinimumOSVersion` that is wildly high to ensure it can't actually be used.

Part of Electric-Coin-Company/zashi-ios#1166.

### DIFF
--- a/support/platform-Info.plist
+++ b/support/platform-Info.plist
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleExecutable</key>
+	<string>libzcashlc</string>
+	<key>CFBundleIdentifier</key>
+	<string>libzcashlc</string>
+	<key>CFBundleName</key>
+	<string>libzcashlc</string>
+	<key>CFBundlePackageType</key>
+	<string>BNDL</string>
+	<key>CFBundleVersion</key>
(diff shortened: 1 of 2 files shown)


support/platform-Info.plist
```diff
@@ -0,0 +1,22 @@
...
+	<key>CFBundleExecutable</key>
+	<string>libzcashlc</string>
+	<key>CFBundleIdentifier</key>
+	<string>libzcashlc</string>
+	<key>CFBundleName</key>
+	<string>libzcashlc</string>
+	<key>CFBundlePackageType</key>
+	<string>BNDL</string>
+	<key>CFBundleVersion</key>
+	<string>0.7.5</string>
```
nuttycom, line 14: This is very likely to bitrot; we should either set it to something like 0.0.1 (since it shouldn't matter) or generate this file?

str4d: Given that the point of this file is to be ignored, I don't think it actually matters if it bitrots? But the problem is I don't actually know if _once we tag the release_ these need to match or not (note that there is no version information anywhere for revision-based dependencies; the `rust/Cargo.toml` version isn't used anywhere, and has been missed in release PRs multiple times in the past).

str4d: Note that I can't generate this file, because the only version I have to go off for that is `rust/Cargo.toml` (which per above has been missed in multiple releases), while what matters for the release is the tag (which cannot be applied before generating).

nuttycom (approved):
simulator works with this ACK

PR #142, opened 2024-06-03
Add FFI methods for fetching exchange rates over Tor

str4d:
We fetch exchange rates on the Rust side of the FFI instead of exposing a direct URL GET method, in order to enable fetching from multiple sources simultaneously to make better use of Arti's async nature.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -32,7 +32,7 @@ dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.35",
 ]
 
 [[package]]
@@ -58,9 +58,9 @@ checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
 
 [[package]]
 name = "arrayref"
-version = "0.3.7"
+version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b4930d2cb77ce62f89ee5d5289b4ac049559b1c45539271f5ed4fdc7db34545"
+checksum = "9d151e35f61089500b617991b791fc8bfd237ae50cd5950803758a179b41e67a"
 
 [[package]]
 name = "arrayvec"
@@ -127,7 +127,7 @@ dependencies = [
  "bitflags 2.6.0",
  "cexpr",
  "clang-sys",
- "itertools",
+ "itertools 0.12.1",
  "lazy_static",
  "lazycell",
  "log",
@@ -137,15 +137,15 @@ dependencies = [
  "regex",
  "rustc-hash",
  "shlex",
- "syn 2.0.68",
+ "syn 2.0.72",
  "which",
 ]
 
 [[package]]
 name = "bip32"
-version = "0.5.1"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7e141fb0f8be1c7b45887af94c88b182472b57c96b56773250ae00cd6a14a164"
+checksum = "aa13fae8b6255872fd86f7faf4b41168661d7d78609f7bfe6771b85c6739a15b"
 dependencies = [
  "bs58",
  "hmac",
@@ -243,9 +243,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.6.0"
+version = "1.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "514de17de45fdb8dc022b1a7975556c53c86f9f0aa5f534b98977b171857c2c9"
+checksum = "a12916984aab3fa6e39d655a33e09c0071eb36d6ab3aea5c2d78551f1df6d952"
 
 [[package]]
 name = "cbc"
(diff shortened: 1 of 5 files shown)


rust/src/tor.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        let mut config_builder = TorClientConfigBuilder::from_directories(
+            tor_dir.join("arti-data"),
+            tor_dir.join("arti-cache"),
+        );
+        config_builder
+            .storage()
+            .permissions()
+            .dangerously_trust_everyone();
```
str4d, line 70: The permissions here need testing. I was having trouble getting Arti to work in the iOS simulator, and one of the things I did to get around it was enable this setting. I eventually also figured out that on the Swift side we needed to pass over `URL.path` instead of `URL.absoluteString`, but I haven't gone back and experimented with other permissions here yet.

rust/Cargo.lock
```diff
@@ -431,1318 +831,3758 @@ dependencies = [
...
+checksum = "3fdb12b2476b595f9358c5161aa467c2438859caa136dec86c26fdd2efe17b92"
+
+[[package]]
+name = "opaque-debug"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c08d65885ee38876c4f86fa503fb49d7b507c2b62552df7c70b2fce627e06381"
+
+[[package]]
+name = "openssl"
```
daira, line 2309: Hmm. OpenSSL still has [frequent and often serious security vulnerabilities](https://openssl.org/news/vulnerabilities.html). These include some bugs that cause platform-dependent and context-dependent undefined behaviour, so won't necessarily show up in testing; for example, the bug with [not restoring XMM registers](https://openssl.org/news/secadv/20230908.txt) on x86-64 processors when using AVX512-IFMA instructions in the Poly1305 code last September. (As it happens that only affected Windows, but that's not the point.)

How is Arti depending on OpenSSL precisely? `tor-llcrypto` does not depend on OpenSSL. At first glance, it seems that [tor-rtcompat](https://crates.io/crates/tor-rtcompat) is depending on it via native-tls. But that should be an optional dependency (only needed for `tokio::TokioNativeTlsRuntime`) I think.

It appears that tor-rtcompat has a [native-tls](https://gitlab.torproject.org/tpo/core/arti/-/blob/main/crates/tor-rtcompat/Cargo.toml?ref_type=heads#L28) feature. Can we disable that?

rust/Cargo.lock
```diff
@@ -431,1318 +831,3758 @@ dependencies = [
...
+ "ctr",
+ "curve25519-dalek",
+ "derive_more",
+ "digest 0.10.7",
+ "ed25519-dalek",
+ "getrandom 0.2.15",
+ "hex",
+ "rand_core 0.5.1",
+ "rand_core 0.6.4",
+ "rsa",
```
daira, line 4361: Is there any way we can force Tor to negotiate Ed25519 rather than 1024-bit RSA? The latter is arguably insecure.
https://github.com/torproject/torspec/blob/8961bb4d83fccb2b987f9899ca83aa430f84ab0c/control-spec.txt#L1760-L1775

daira: Hmm, I guess [RSA is not actually relied on](https://spec.torproject.org/tor-spec/relay-keys.html#identity).

rust/Cargo.lock
str4d: Arti does not; we can configure it to use Rustls / ring. Note however that this openssl dependency is only actually used on Linux; on iOS it will be using the built-in Security framework.

str4d: I looked into this, and in order to keep the `ring` version unified, we would need to downgrade from `arti-client 0.11` (which is already a downgrade for `ring` unification) to `arti-client 0.9.0` (and pin a bunch of dependencies to specifically `0.9.0` etc, as Arti migrated to `rustls 0.21` in `0.9.1`).

I've pushed this in ce10808fa1475de0d289d723d9fe2ac2d69f7b02 for you to consider. It still needs testing to confirm it works.

nuttycom (approved):
utACK d0907c1c3cc147287bcd2a4ab9805575e7771bdf

rust/src/tor/exchange_rate.rs
```diff
@@ -0,0 +1,143 @@
+use futures_util::{join, FutureExt};
+use rand::{seq::IteratorRandom, thread_rng};
+use serde::{Deserialize, Deserializer};
+use tracing::{error, trace};
+
+use super::Error;
+
+impl super::Client {
+    #[tracing::instrument(skip(self))]
+    pub(crate) async fn get_exchange_rate_usd(&self) -> Result<f64, Error> {
```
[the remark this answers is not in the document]
str4d: Now that we require at least 3 data sources and we continue to take the median, we have pretty good defence against individual source divergence.

pacu (approved):

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-light-client-ffi/compare/d0907c1c3cc147287bcd2a4ab9805575e7771bdf..2516a94f8bdc540d951c38b66e9c07e2b8c29cb4) to pull in the latest Rust crate changes (which include ZIP 320 bugfixes), and do some other updates.

rust/src/tor.rs
str4d: This was removed after we moved this code to `zcash_client_backend`, after I confirmed it wasn't necessary.

daira (commented):
Post-hoc ACK

PR #148, opened 2024-08-14
Add FFI bindings for `WalletRead::transaction_data_requests` and `WalletWrite::set_transaction_status`

nuttycom:


rust/src/lib.rs
```diff
@@ -2662,7 +2662,19 @@ pub unsafe extern "C" fn zcashlc_decrypt_and_store_transaction(
...
+        let mined_height = if mined_height > 0 {
+            let h = u32::try_from(mined_height)
+                .map_err(|e| anyhow!("Block height outside valid range: {}", e))?;
+            Some(h.into())
+        } else {
+            // We do not provide a mined height to `decrypt_and_store_transaction` for either
+            // transactions in the mempool or for transactions that have been mined on a fork
+            // but not in the main chain.
+            None
+        };
```
str4d, line 2677: Missing comment at the top.

Also, in the equivalent Android SDK change I handled this on the Kotlin side ([gRPC parsing](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/pull/1559/files#diff-af6f9221fb596338351cd262fe59664d5b2114ee07cabbd90de13864244991d5R33-R40), [conversion to main Kotlin type](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/pull/1559/files#diff-ccf0b549a7923796d7fd0266c843dc7264fa30f3a6b168c5106f0c9164021e00R30-R34)), and used `-1` separately as a sentinel for `None` across the FFI ([Kotlin side](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/pull/1559/files#diff-df1ff90ff1acc078d2f9983b61bc2bd21e6137845526b00c41d558df9358daccR315), [Rust side](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/pull/1559/files#diff-63f3f4a8b98989699289a769aae68d9a8671559608f0119647df4a07194fcab3R1658)).

rust/src/lib.rs
```diff
@@ -3173,6 +3174,69 @@ pub unsafe extern "C" fn zcashlc_create_proposed_transactions(
...
+    /// The requested transaction ID was not recognized by the node.
+    TxidNotRecognized,
+    /// The requested transaction ID corresponds to a transaction that is recognized by the node,
+    /// but is in the mempool or is otherwise not mined in the main chain (but may have been mined
+    /// on a fork that was reorged away).
+    NotInMainChain,
+    /// The requested transaction ID corresponds to a transaction that has been included in the
+    /// block at the provided height.
+    Mined(u32),
+}
```
str4d, line 3190: By comparison, in [the Android SDK](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/pull/1559/files#diff-63f3f4a8b98989699289a769aae68d9a8671559608f0119647df4a07194fcab3R1686-R1690) I passed sentinel values across because JNI doesn't have an enum type I could use. This is much nicer.

rust/src/lib.rs
```diff
@@ -3173,6 +3174,69 @@ pub unsafe extern "C" fn zcashlc_create_proposed_transactions(
...
+    status: FfiTransactionStatus,
+) {
+    let res = catch_panic(|| {
+        let network = parse_network(network_id)?;
+        let mut db_data = unsafe { wallet_db(db_data, db_data_len, network)? };
+
+        let txid_bytes = unsafe { slice::from_raw_parts(txid_bytes, txid_bytes_len) };
+        let mut txid = [0u8; 32];
+        txid.copy_from_slice(txid_bytes);
+        let txid = TxId::from_bytes(txid);
```
str4d, line 3225: This will panic, which gets caught but with probably a not great error. If you instead use:
```suggestion
        let txid = TxId::read(&txid_bytes[..])?;
```
then you get a nicer error.

rust/src/lib.rs
```diff
@@ -3237,6 +3238,156 @@ pub unsafe extern "C" fn zcashlc_set_transaction_status(
...
+        let (ptr, len) = ptr_from_vec(v);
+        Box::into_raw(Box::new(FfiTransactionDataRequests { ptr, len }))
+    }
+}
+
+/// Frees an array of FfiTransactionDataRequest values as allocated by `zcashlc_transaction_data_requests`.
+///
+/// # Safety
+///
+/// - `ptr` must be non-null and must point to a struct having the layout of [`FfiTransactionDataRequests`].
```
str4d, line 3324: Technically we allow `ptr` to be null, but if it is non-null it must...

str4d (approved):
utACK 64ec3b65c985f56bdab9c8bda1476c419cdd456f with suggestions.

rust/src/lib.rs
nuttycom:
```suggestion
/// - `ptr` if `ptr` is non-null it must point to a struct having the layout of [`FfiTransactionDataRequests`].
```

rust/src/lib.rs
```diff
@@ -2662,7 +2662,19 @@ pub unsafe extern "C" fn zcashlc_decrypt_and_store_transaction(
         //   from their encoding.
         let tx = Transaction::read(tx_bytes, BranchId::Sapling)?;
 
-        match decrypt_and_store_transaction(&network, &mut db_data, &tx) {
+        //
```
nuttycom, line 2667:
```suggestion
        // Following the conventions of the `zcashd` `getrawtransaction` RPC method,
        // negative values (specifically -1) indicate that the transaction may have been 
        // mined, but in a fork of the chain rather than the main chain, whereas a value 
        // of zero indicates that the transaction is in the mempool. We do not distinguish
        // between these in `librustzcash`, and so both cases are mapped to `None`,
        // indicating that the mined height of the transaction is simply unknown.

rust/src/lib.rs
nuttycom: Since we don't have a representation for the distinction between "in the mempool" and "otherwise mined in a fork" in `librustzcash` yet, I decided to allow the caller to directly propagate the response of the RPC call if they wanted; basically the only case we're currently interested in on the Rust side is when the mined height is known, all the other cases just get mapped to `None`. In the future if `librustzcash` begins to distinguish between these cases by using a bespoke enum instead of an `Option`, we'll update the FFI at that point.

PR #151, opened 2024-09-27
Update `zcash_client_sqlite` to fix potential note commitment tree corruption.

nuttycom:
As part of this change, the `zcashlc_get_nearest_rewind_height` FFI method is removed. The return type of `zcashlc_rewind_to_height` has been changed to an `i64` that represents the height to which the database was actually rewound, or `-1` if the rewind was unsuccessful. Additionally, `zcashlc_rewind_to_height` now takes an output parameter that the Rust backend can use to set the minimum valid rewind height, if any.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- `zcashlc_rewind_to_height` now returns an `i64` value instead of a boolean. The
+  value `-1` indicates failure; any other height indicates the height to which the
+  data store was actually truncated. Also, this procedure now takes an additional
+  `safe_rewind_ret` parameter that, on failure to rewind, will be set to the
+  minimum height for which the rewind would succeed or will remain unmodified if
+  no such height can be determined.
+
+### Removed
+- `zcashlc_get_nearest_rewind_height` has been removed. The return value of 
+  `zcashlc_rewind_to_height`, or in the case of rewind failure the value of its
+  `safe_rewind_ret` return parameter should be used instead.
+
+### Fixed
+- This release fixes a potential source of corruption in wallet note commitment
+  trees related to incorrect handling of chain reorgs. It includes a database
+  migration that will repair the corrupted database state of any wallet
+  affected by this corner case.
+
 ## 0.9.1 - 2024-08-21
 
 ### Fixed
--- a/releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
+++ b/releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
@@ -1008,8 +1008,17 @@ bool zcashlc_seed_fingerprint(const uint8_t *seed,
                               uint8_t *signature_bytes_ret);
 
 /**
- * Returns the most recent block height to which it is possible to reset the state
- * of the data database.
+ * Rewinds the data database to at most the given height.
+ *
+ * If the requested height is greater than or equal to the height of the last scanned block, this
+ * function does nothing.
+ *
+ * This procedure returns the height to which the database was actually rewound, or `-1` if no
+ * rewind was performed.
+ *
+ * If the requested rewind could not be performed, but a rewind to a different (greater) height
+ * would be valid, the `safe_rewind_ret` output parameter will be set to that value on completion;
+ * otherwise, it will remain unmodified.
  *
  * # Safety
  *
@@ -1020,30 +1029,11 @@ bool zcashlc_seed_fingerprint(const uint8_t *seed,
  * - The total size `db_data_len` must be no larger than `isize::MAX`. See the safety
  *   documentation of pointer::offset.
  */
-int32_t zcashlc_get_nearest_rewind_height(const uint8_t *db_data,
-                                          uintptr_t db_data_len,
-                                          int32_t height,
-                                          uint32_t network_id);
-
-/**
- * Rewinds the data database to the given height.
- *
- * If the requested height is greater than or equal to the height of the last scanned
- * block, this function does nothing.
- *
- * # Safety
- *
- * - `db_data` must be non-null and valid for reads for `db_data_len` bytes, and it must have an
- *   alignment of `1`. Its contents must be a string representing a valid system path in the
- *   operating system's preferred representation.
- * - The memory referenced by `db_data` must not be mutated for the duration of the function call.
- * - The total size `db_data_len` must be no larger than `isize::MAX`. See the safety
- *   documentation of pointer::offset.
- */
-bool zcashlc_rewind_to_height(const uint8_t *db_data,
-                              uintptr_t db_data_len,
-                              int32_t height,
-                              uint32_t network_id);
+int64_t zcashlc_rewind_to_height(const uint8_t *db_data,
+                                 uintptr_t db_data_len,
+                                 uint32_t height,
+                                 uint32_t network_id,
+                                 uint32_t *safe_rewind_ret);
 
 /**
  * Adds a sequence of Sapling subtree roots to the data store.
--- a/releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
+++ b/releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
@@ -1011,17 +1011,18 @@ bool zcashlc_seed_fingerprint(const uint8_t *seed,
  * Rewinds the data database to at most the given height.
  *
  * If the requested height is greater than or equal to the height of the last scanned block, this
- * function does nothing.
+ * function sets the `safe_rewind_ret` output parameter to `-1` and does nothing else.
  *
  * This procedure returns the height to which the database was actually rewound, or `-1` if no
  * rewind was performed.
  *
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,25 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- `zcashlc_rewind_to_height` now returns an `i64` value instead of a boolean. The
+  value `-1` indicates failure; any other height indicates the height to which the
+  data store was actually truncated. Also, this procedure now takes an additional
+  `safe_rewind_ret` parameter that, on failure to rewind, will be set to the
+  minimum height for which the rewind would succeed or will remain unmodified if
+  no such height can be determined.
```
daira, line 15: I generally think that it's a bad idea for out parameters to only be set conditionally. It's too easy to fail to initialize them in the caller (for example, [here](https://github.com/zcash/zcash/pull/6926/files#diff-63c7710d72381cffe692175dd70e47f1891bf75c3e09249ed826f9df3917d02dL321) there was undefined behaviour due to `nFeeRequired` being uninitialized and only conditionally set). In this case I think it would be better to use an out parameter that is a `std::optional`, and set it to `std::nullopt` at the start of `zcashlc_rewind_to_height` (suggestion below).

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
```diff
@@ -1008,28 +1008,17 @@ bool zcashlc_seed_fingerprint(const uint8_t *seed,
                               uint8_t *signature_bytes_ret);
 
 /**
- * Returns the most recent block height to which it is possible to reset the state
- * of the data database.
+ * Rewinds the data database to at most the given height.
  *
- * # Safety
+ * If the requested height is greater than or equal to the height of the last scanned block, this
+ * function does nothing.
```
daira, line 1014:
```suggestion
 * function sets `*safe_rewind_ret` to `-1` and does nothing else.
```

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
```diff
@@ -1008,28 +1008,17 @@ bool zcashlc_seed_fingerprint(const uint8_t *seed,
...
-/**
- * Rewinds the data database to the given height.
+ * This procedure returns the height to which the database was actually rewound, or `-1` if no
+ * rewind was performed.
  *
- * If the requested height is greater than or equal to the height of the last scanned
- * block, this function does nothing.
+ * If the requested rewind could not be performed, but a rewind to a different (greater) height
+ * would be valid, the `safe_rewind_ret` output parameter will be set to that value on completion;
+ * otherwise, it will remain unmodified.
```
daira, line 1021:
```suggestion
 * otherwise, it will be set to `-1`.
```

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
```diff
@@ -1040,10 +1029,11 @@ int32_t zcashlc_get_nearest_rewind_height(const uint8_t *db_data,
  * - The total size `db_data_len` must be no larger than `isize::MAX`. See the safety
  *   documentation of pointer::offset.
  */
```
daira, line 1031:
```suggestion
 * - `safe_rewind_ret` must be non-null, aligned, and valid for writing a `uint32_t`.
 */
```

rust/src/lib.rs
```diff
@@ -1606,61 +1607,17 @@ pub unsafe extern "C" fn zcashlc_seed_fingerprint(
     unwrap_exc_or(res, false)
 }
 
-/// Returns the most recent block height to which it is possible to reset the state
-/// of the data database.
+/// Rewinds the data database to at most the given height.
 ///
-/// # Safety
+/// If the requested height is greater than or equal to the height of the last scanned block, this
+/// function does nothing.
```
daira, line 1613:
```suggestion
 * function sets `*safe_rewind_ret` to `-1` and does nothing else.
```

rust/src/lib.rs
```diff
@@ -1606,61 +1607,17 @@ pub unsafe extern "C" fn zcashlc_seed_fingerprint(
...
-
-/// Rewinds the data database to the given height.
+/// This procedure returns the height to which the database was actually rewound, or `-1` if no
+/// rewind was performed.
 ///
-/// If the requested height is greater than or equal to the height of the last scanned
-/// block, this function does nothing.
+/// If the requested rewind could not be performed, but a rewind to a different (greater) height
+/// would be valid, the `safe_rewind_ret` output parameter will be set to that value on completion;
+/// otherwise, it will remain unmodified.
```
daira, line 1620:
```suggestion
 * otherwise, it will be set to `-1`.
```

rust/src/lib.rs
```diff
@@ -1606,61 +1607,17 @@ pub unsafe extern "C" fn zcashlc_seed_fingerprint(
...
 ///
 /// # Safety
 ///
```
daira, line 1623:
```suggestion
///
/// - `safe_rewind_ret` must be non-null, aligned, and valid for writing a `uint32_t`.
```

rust/src/lib.rs
```diff
@@ -1674,20 +1631,36 @@ pub unsafe extern "C" fn zcashlc_get_nearest_rewind_height(
 pub unsafe extern "C" fn zcashlc_rewind_to_height(
     db_data: *const u8,
     db_data_len: usize,
-    height: i32,
+    height: u32,
     network_id: u32,
-) -> bool {
+    safe_rewind_ret: *mut u32,
+) -> i64 {
     let res = catch_panic(|| {
```
daira, line 1638:
```suggestion
    unsafe { *safe_rewind_ret = -1; }
    let res = catch_panic(|| {
```

rust/src/lib.rs
```diff
@@ -1674,20 +1631,36 @@ pub unsafe extern "C" fn zcashlc_get_nearest_rewind_height(
 pub unsafe extern "C" fn zcashlc_rewind_to_height(
     db_data: *const u8,
     db_data_len: usize,
-    height: i32,
+    height: u32,
     network_id: u32,
-) -> bool {
+    safe_rewind_ret: *mut u32,
```
nuttycom, line 1636:
```suggestion
    safe_rewind_ret: *mut i64,
```

rust/src/lib.rs
```diff
@@ -1674,20 +1631,36 @@ pub unsafe extern "C" fn zcashlc_get_nearest_rewind_height(
...
+        let height = BlockHeight::from(height);
+        let result_height = db_data.truncate_to_height(height);
+
+        result_height.map_or_else(
+            |err| match err {
+                SqliteClientError::RequestedRewindInvalid {
+                    safe_rewind_height: Some(h),
+                    ..
+                } => {
+                    unsafe { *safe_rewind_ret = h.into() };
```
nuttycom, line 1651:
```suggestion
                    unsafe { *safe_rewind_ret = u32::from(h).into() };
```

rust/src/lib.rs
```diff
@@ -1606,64 +1607,21 @@ pub unsafe extern "C" fn zcashlc_seed_fingerprint(
     unwrap_exc_or(res, false)
 }
 
-/// Returns the most recent block height to which it is possible to reset the state
-/// of the data database.
+/// Rewinds the data database to at most the given height.
 ///
-/// # Safety
+/// If the requested height is greater than or equal to the height of the last scanned block, this
+ * function sets the `safe_rewind_ret` output parameter to `-1` and does nothing else.
```
daira, line 1613:
```suggestion
/// function sets the `safe_rewind_ret` output parameter to `-1` and does nothing else.
```
Sorry, copy-paste error in my previous suggestion.

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
daira, line 1: I did not review this file. Generated files shouldn't be checked in! It's a security hazard. Ditto for the next two files.

daira (approved):
utACK

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
nuttycom: I invite you to read this garbage:
```
/// - Parameters:
///     - name: The name of the package, or nil to deduce it from the URL.
///     - url: The valid Git URL of the package.
///     - version: The minimum version requirement.
static func package(url: String, from version: Version) -> Package.Dependency
```
The _only_ option that Swift provides for remote package download is a git URL. Otherwise, your only option is a local path. 

Swift Package Manager literally does not work for anything but files stored in git.

It takes ~20 minutes on a M1 macbook to build our native binary packages. This is why Pacu wanted them to be an external dependency in the first place, so that the build cost could be incurred once and then the resulting artifacts reused. But the only option Swift supports is to store them in git. It is just as stupid as it sounds.

nuttycom:
Subsumed by #154

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
daira: 😭

PR #154, opened 2024-10-04
Release zcash-light-client-ffi version 0.10.0

nuttycom:


CHANGELOG.md
```diff
@@ -6,10 +6,31 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## Unreleased
 
+## 0.10.0 - 2024-10-04
+
+### Changed
+- `zcashlc_rewind_to_height` now returns an `i64` value instead of a boolean. The
+  value `-1` indicates failure; any other height indicates the height to which the
+  data store was actually truncated. Also, this procedure now takes an additional
+  `safe_rewind_ret` parameter that, on failure to rewind, will be set to the
+  minimum height for which the rewind would succeed or will remain unmodified if
```
daira, line 16:
```suggestion
  minimum height for which the rewind would succeed, or to -1 if
```

rust/src/lib.rs
```diff
@@ -1606,64 +1607,21 @@ pub unsafe extern "C" fn zcashlc_seed_fingerprint(
...
 ///
-/// If the requested height is greater than or equal to the height of the last scanned
-/// block, this function does nothing.
+/// If the requested rewind could not be performed, but a rewind to a different (greater) height
+/// would be valid, the `safe_rewind_ret` output parameter will be set to that value on completion;
+/// otherwise, it will be set to `-1`.
 ///
 /// # Safety
 ///
+/// - `safe_rewind_ret` must be non-null, aligned, and valid for writing a `uint32_t`.
```
daira, line 1624:
```suggestion
/// - `safe_rewind_ret` must be non-null, aligned, and valid for writing an `int64_t`.
```
Also move this to the bottom of the doc comment to match the parameter order.

rust/src/lib.rs
```diff
@@ -2080,11 +2057,18 @@ impl FfiWalletSummary {
...
-        let scan_progress = if let Some(progress) = summary.scan_progress() {
-            Box::into_raw(Box::new(FfiScanProgress {
-                numerator: *progress.numerator(),
-                denominator: *progress.denominator(),
-            }))
+        let scan_progress = if let Some(scan_progress) = summary.scan_progress() {
+            if let Some(recovery_progress) = summary.recovery_progress() {
+                Box::into_raw(Box::new(FfiScanProgress {
+                    numerator: *scan_progress.numerator() + *recovery_progress.numerator(),
+                    denominator: *scan_progress.denominator() + *recovery_progress.denominator(),
```
daira, line 2064: This produces strange results if the `scan_progress` and `recovery_progress` denominators are different. I don't know what is best here, but if you want the two progress ratios to be equally weighted, then it would be `numerator: *scan_progress.numerator * *recovery_progress.denominator + *recovery_progress.numerator * *scan_progress.denominator, denominator: 2 * *scan_progress.denominator * *recovery_progress.denominator`.

daira (approved):
utACK modulo comments.

rust/src/lib.rs
nuttycom: In this particular case, these ratios are unreduced with the denominators being the number of notes in the scan range. So directly adding the numerators and denominators here is fine; this is going to change soon anyway once I rework the types as I suggest in https://github.com/zcash/librustzcash/issues/1559

daira:
Post-hoc ACK be1575f8d2dce5b7502448ef1f8619e2ca9359d9

PR #158, opened 2024-10-23
Add ZIP 32 Arbitrary Key derivation

str4d:


rust/src/derivation.rs
```diff
@@ -443,3 +444,96 @@ pub unsafe extern "C" fn zcashlc_get_sapling_receiver_for_unified_address(
...
+    context_string: *const u8,
+    context_string_len: usize,
+    seed: *const u8,
+    seed_len: usize,
+) -> *mut FfiBoxedSlice {
+    let res = catch_panic(|| {
+        let context_string = unsafe { slice::from_raw_parts(context_string, context_string_len) };
+        let seed = unsafe { slice::from_raw_parts(seed, seed_len) };
+
+        let key = arbitrary::SecretKey::from_path(context_string, seed, &[]);
```
nuttycom, line 484: Given that this will panic if `context_string` is greater than 252 bytes, should that be checked?

Also, now that I'm thinking about this, why was this implemented as a panic in https://github.com/zcash/zip32/pull/19/files?

nuttycom (approved):
utACK 28c4ac50dfd6952fba1eff170142dafb4d753dec with comment/retroactive question about the `zip32::arbitrary::SecretKey::from_path`.

rust/src/derivation.rs
nuttycom: I suppose that this is not intended to be a dynamically generated value, so a panic makes sense, because you should be using a constant for the context string, but generally it seems better to use `Result` for this sort of thing, because the caller can always choose to `unwrap`. A user who simply uses e.g. IDE autocomplete to take advantage of a method like `arbitrary::SecretKey::from_path` will miss that it can panic, and this can therefore be a source of avoidable runtime crashes.

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+///
+/// See the following sections of ZIP 316 for details on how to interpret typecodes:
+/// - [List of known typecodes](https://zips.z.cash/zip-0316#encoding-of-unified-addresses)
+/// - [Adding new types](https://zips.z.cash/zip-0316#adding-new-types)
+/// - [Metadata Items](https://zips.z.cash/zip-0316#metadata-items)
+///
+/// # Safety
+///
+/// - `ua` must be non-null and must point to a null-terminated UTF-8 string containing an
+///   encoded Unified Address.
```
daira, line 177:
```suggestion
///   encoded Unified Address.
/// - The memory referenced by `ua` must not be mutated for the duration of the function call.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+
+/// Derives and returns a unified spending key from the given seed for the given account ID.
+///
+/// Returns the binary encoding of the spending key. The caller should manage the memory of (and
+/// store, if necessary) the returned spending key in a secure fashion.
+///
+/// # Safety
+///
+/// - `seed` must be non-null and valid for reads for `seed_len` bytes, and it must have an
+///   alignment of `1`.
```
daira, line 301: Having an alignment of 1 imposes no requirement (and you would not expect it to have any other alignment, having type `*const u8`).

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+/// - The memory referenced by `seed` must not be mutated for the duration of the function call.
+/// - The total size `seed_len` must be no larger than `isize::MAX`. See the safety documentation
+///   of pointer::offset.
```
daira, line 304:
```suggestion
///   of `pointer::offset`.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+/// and returns the resulting encoded UFVK string. `usk_ptr` should point to an array of `usk_len`
+/// bytes containing a unified spending key encoded as returned from the `zcashlc_create_account`
+/// or `zcashlc_derive_spending_key` functions.
+///
+/// # Safety
+///
+/// - `usk_ptr` must be non-null and must point to an array of `usk_len` bytes.
+/// - The memory referenced by `usk_ptr` must not be mutated for the duration of the function call.
+/// - The total size `usk_len` must be no larger than `isize::MAX`. See the safety documentation
+///   of pointer::offset.
```
daira, line 339:
```suggestion
///   of `pointer::offset`.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+            .map(UnifiedAddressParser)
+            .map_err(|e| anyhow!("Invalid Unified Address: {}", e).into())
+    }
+}
+
+/// Returns the transparent receiver within the given Unified Address, if any.
+///
+/// # Safety
+///
+/// - `ua` must be non-null and must point to a null-terminated UTF-8 string.
```
daira, line 376:
```suggestion
/// - `ua` must be non-null and must point to a null-terminated UTF-8 string.
/// - The memory referenced by `ua` must not be mutated for the duration of the function call.
```

rust/src/derivation.rs
daira, line 177: Is that a safety requirement, or a correctness requirement? I think it's a correctness requirement, and a null pointer will be returned if it is not a correct UA.
```suggestion
/// - `ua` must be non-null and must point to a null-terminated UTF-8 string.
/// - The memory referenced by `ua` must not be mutated for the duration of the function call.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+        }
+    });
+    unwrap_exc_or_null(res)
+}
+
+/// Returns the Sapling receiver within the given Unified Address, if any.
+///
+/// # Safety
+///
+/// - `ua` must be non-null and must point to a null-terminated UTF-8 string.
```
daira, line 417:
```suggestion
/// - `ua` must be non-null and must point to a null-terminated UTF-8 string.
/// - The memory referenced by `ua` must not be mutated for the duration of the function call.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+/// mainnet, and so on). You can think of it as a context-specific seed fingerprint that can be used
+/// as (static) key material.
+///
+/// `context_string` is a globally-unique non-empty sequence of at most 252 bytes that identifies
+/// the desired context.
+///
+/// # Safety
+///
+/// - `context_string` must be non-null and valid for reads for `context_string_len` bytes, and it
+///   must have an alignment of `1`.
```
daira, line 461:
```suggestion
/// - `context_string` must be non-null and valid for reads for `context_string_len` bytes.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+/// - The memory referenced by `context_string` must not be mutated for the duration of the function
+///   call.
+/// - The total size `context_string_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of pointer::offset.
```
daira, line 465:
```suggestion
///   documentation of `pointer::offset`.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+/// - `seed` must be non-null and valid for reads for `seed_len` bytes, and it must have an
+///   alignment of `1`.
+/// - The memory referenced by `seed` must not be mutated for the duration of the function call.
+/// - The total size `seed_len` must be no larger than `isize::MAX`. See the safety documentation
+///   of pointer::offset.
```
daira, line 470:
```suggestion
///   of `pointer::offset`.
```

rust/src/derivation.rs
daira: If it causes a panic then it must be documented in a Panics section of the doc comment. I think it is preferable to have it be an error, though.

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+
+/// Derives and returns a ZIP 32 Arbitrary Key from the given seed at the account level.
+///
+/// `context_string` is a globally-unique non-empty sequence of at most 252 bytes that identifies
+/// the desired context.
+///
+/// # Safety
+///
+/// - `context_string` must be non-null and valid for reads for `context_string_len` bytes, and it
+///   must have an alignment of `1`.
```
daira, line 499:
```suggestion
/// - `context_string` must be non-null and valid for reads for `context_string_len` bytes.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+/// - The memory referenced by `context_string` must not be mutated for the duration of the function
+///   call.
+/// - The total size `context_string_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of pointer::offset.
```
daira, line 503:
```suggestion
///   documentation of `pointer::offset`.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+/// - `seed` must be non-null and valid for reads for `seed_len` bytes, and it must have an
+///   alignment of `1`.
```
daira, line 505:
```suggestion
/// - `seed` must be non-null and valid for reads for `seed_len` bytes`.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+/// - The memory referenced by `seed` must not be mutated for the duration of the function call.
+/// - The total size `seed_len` must be no larger than `isize::MAX`. See the safety documentation
+///   of pointer::offset.
```
daira, line 508:
```suggestion
///   of `pointer::offset`.
```

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+        let seed = unsafe { slice::from_raw_parts(seed, seed_len) };
+
+        let key = arbitrary::SecretKey::from_path(context_string, seed, &[]);
+
+        Ok(FfiBoxedSlice::some(key.data().to_vec()))
+    });
+    unwrap_exc_or_null(res)
+}
+
+/// Derives and returns a ZIP 32 Arbitrary Key from the given seed at the account level.
```
daira, line 491: Should this document what the path is?

rust/src/derivation.rs
```diff
@@ -0,0 +1,539 @@
...
+///
+/// `context_string` is a globally-unique non-empty sequence of at most 252 bytes that identifies
+/// the desired context.
```
daira, line 494: Similar to `zcashlc_derive_arbitrary_wallet_key`, either document the panic or make it an error.

daira (approved):
utACK with comments. The comments about either documenting panics or making them into errors (I prefer the latter) are blocking. If the decision is to keep the panics, that would also have to be documented at the SDK layer (e.g. https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/pull/1501/files#r1823372962).

rust/src/derivation.rs
str4d: This does not cause a panic in the Swift SDK. There's a `catch_panic` a few lines above here, that converts the Rust panic into an error.

daira (approved):
utACK

PR #161, opened 2024-11-04
Update build artifacts for zip32 arbitrary keys functionality.

nuttycom:


str4d:
This was a temporary branch for manual preview binaries. We now review and merge proposed changes into release branches, from which preview binaries are automatically built.

PR #162, opened 2024-11-05
CI: Add workflow to build or update binary previews

str4d:


.github/workflows/build-binary-preview.yml
```diff
@@ -0,0 +1,30 @@
+name: Binary preview
+
+on:
+  push:
+    branches:
+      - 'ffi-**'
+      - 'hotfix-**'
```
nuttycom, line 7: Mild personal preference:
```suggestion
      - 'ffi/**'
      - 'hotfix/**'
```

nuttycom: Something that I have wanted for a while is to slightly generalize the FFI and the build process here so that I can use this API directly from the Haskell FFI; at present, I have to duplicate a lot of the code. In the case that we eventually do that, the build _artifacts_ would be specifically for use from Swift, but the code would be reusable. If/when we use local builds for Swift, as you've suggested, I'd prefer that this repository be included in `zcash-swift-wallet-sdk` as a submodule instead of moving it in wholesale.

Not directly related to the choice of branch prefixes, but slightly relevant as the artifacts are the primary bit that's specific to use from Swift; I don't think we rely on any APIs that are (i/Mac)OS specific?

str4d: We do rely on Apple-specific APIs, for logging and tracing: https://github.com/Electric-Coin-Company/zcash-light-client-ffi/blob/main/rust/wrapper.c

They are however gated off behind target flags. This was only done so I could develop the FFI backend on non-Apple machines, but it has the nice side-effect of cutting them out of non-Apple builds.

> I'd prefer that this repository be included in zcash-swift-wallet-sdk as a submodule instead of moving it in wholesale.

We will never do that. This repository is tarnished with the history of having large binaries committed to it, and including it as a submodule would carry that forward indefinitely (along with all of the corrupted checkout problems that I have suffered through). When we move back to local builds for Swift, this repository will be archived and frozen.

If we _separately_ are motivated to turn this C FFI into a resource backing more than one front-end SDK (which has never been a design consideration for us before), then it should be as part of a brand new repository that does not have the very weighty baggage of this repository. However, if we're doing that (supporting multiple front-end SDKs) then we would also want to switch to a generated FFI layer.

In any case, I picked the above branch prefixes because that's what was historically used here (`hotfix-**` in particular was used by you two weeks ago in #157). But I don't care strongly about it, as long as it matches what we're actually doing.

.github/workflows/build-binary-preview.yml
```diff
@@ -0,0 +1,30 @@
...
+  build:
+    name: Build
+    runs-on: macos-12
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Create, or merge into, the corresponding preview branch
+        run: git switch -c '${{ github.ref_name }}-preview' || git switch -m '${{ github.ref_name }}-preview'
```
str4d, line 19: Gah, it only occurred to me after I clicked the merge button that @nuttycom's branch name suggestion also needed to be changed here:
```suggestion
        run: git switch -c '${{ github.ref_name }}/preview' || git switch -m '${{ github.ref_name }}/preview'
```
Otherwise the `preview` suffix gets interpreted as part of the version.

str4d: Fixed in #163.

PR #163, opened 2024-11-05
CI: Change naming pattern for binary preview branches to avoid bugs

str4d:
- After changing from `-` to `/` logical separation, the `-preview` suffix was mis-interpretable as part of the version.
- Using a `preview` suffix would potentially cause the workflow to recursively loop, because the workflow branch filter would match both the release / hotfix branches, and every preview branch. We now use a `preview` prefix to domain-separate the preview branches.


str4d:
Self-ACK

PR #164, opened 2024-11-05
CI: Install the necessary Rust targets before building binary previews

str4d:


str4d:
Self-ACK.

nuttycom (commented):
post-hoc ACK

PR #165, opened 2024-11-05
CI: Specify branch to commit and push binary previews to

str4d:
This is necessary to get the stefanzweifel/git-auto-commit-action@v5 action to set the upstream branch.


nuttycom (approved):
utACK

PR #166, opened 2024-11-13
Update transaction proposal creation to use `MultiOutputChangeStrategy`

nuttycom:


rust/src/lib.rs
```diff
@@ -2253,17 +2257,22 @@ pub unsafe extern "C" fn zcashlc_decrypt_and_store_transaction(
...
+) {
+    (
+        MultiOutputChangeStrategy::new(
+            StandardFeeRule::Zip317,
+            change_memo,
+            ShieldedProtocol::Orchard,
+            DustOutputPolicy::default(),
+            SplitPolicy::with_min_output_value(
+                NonZeroUsize::new(4).unwrap(),
+                Zatoshis::const_from_u64(10000000),
```
str4d, line 2272: Is 1 ZEC potentially a bit high? Current Zcash price is US$37, and AIUI `MultiOutputChangeStrategy` prefers targeting the minimum value over the target number of notes, so for small wallets (coffee shop usage) this is still going to tend towards one note.

str4d (approved):
utACK fa975715759ea771bb7c46f1686da8adcf611485 with question.

rust/src/lib.rs
nuttycom: This is 0.1 ZEC, right?

str4d: Ah yep, I miscounted. Could we use `1000_0000` to make it clearer?

nuttycom, line 2272:
```suggestion
                Zatoshis::const_from_u64(1000_0000),
```

PR #172, opened 2024-12-04
Use UUIDs instead of 32-bit indexes for account identifiers.

nuttycom:


rust/src/lib.rs
```diff
@@ -310,6 +310,59 @@ pub unsafe extern "C" fn zcashlc_init_data_database(
...
+///
+/// - `ptr` must be non-null and must point to a struct having the layout of [`FfiAccount`].
+#[no_mangle]
+pub unsafe extern "C" fn zcashlc_free_account(ptr: *mut FfiAccount) {
+    if !ptr.is_null() {
+        let account: Box<FfiAccount> = unsafe { Box::from_raw(ptr) };
+        if !(account.account_name).is_null() {
+            unsafe { zcashlc_string_free(account.account_name) }
+        }
+        if !(account.key_source).is_null() {
```
str4d, line 359: What is this parens syntax needed for?

rust/src/lib.rs
```diff
@@ -385,6 +438,42 @@ pub unsafe extern "C" fn zcashlc_list_accounts(
...
+        let db_data = unsafe { wallet_db(db_data, db_data_len, network)? };
+        let account_uuid = account_uuid_from_bytes(account_uuid_bytes)?;
+
+        Ok(if let Some(account) = db_data.get_account(account_uuid)? {
+            Box::into_raw(Box::new(FfiAccount::from_account(&account)))
+        } else {
+            ptr::null_mut()
+        })
+    });
+    unwrap_exc_or_null(res)
```
str4d, line 474: There's a representational problem here. You're returning `null` to mean both `Ok(None)` and `Err(_)`, which potentially means that an error can't be correctly detected on the Swift side.

In other APIs I've resolved this by adding `FfiFoobar::some` and `FfiFoobar::none` methods, with the latter generating a non-null `FfiFoobar` containing sentinel values indicating `None` (e.g. a ZIP 32 account index value that cannot be valid; e.g. cast `u32` to `i32` and then use the negative number space as the sentinel values). In other places, we have a natural wrapping struct that is always non-null that contains a nullable pointer to the optional struct, which also provides a distinguisher.

rust/src/lib.rs
```diff
@@ -542,6 +632,11 @@ pub unsafe extern "C" fn zcashlc_free_ffi_uuid(ptr: *mut FfiUuid) {
 /// Adds a new account to the wallet by importing the UFVK that will be used to detect incoming
 /// payments.
 ///
+/// Derivation metadata may optionally be included. To indicate that no derivation metadata is
+/// available, the `seed_fingerprint` argument should be set to the null pointer and
+/// `hd_account_index` should be set to the value `u32::MAX`. Derivation metadata will not be
+/// stored unless both the seed fingerprint and the HD account index are provided.
```
str4d, line 638: Do we want to just not store it, or should we return an error if this occurs? It would likely indicate a programming error.

str4d (changes requested):
Reviewed c7ddd9da05096cd0e7361b5cf31495ef868e2a3c

rust/src/lib.rs
nuttycom: The compiler was for some reason trying to bind the `!` operator to the pointer, but parenthesizing makes the `is_null` call typecheck.

rust/src/lib.rs
nuttycom: Okay, I wasn't sure of the right thing to do here. I'll change this to a sentinel.

rust/src/lib.rs
nuttycom: An error sounds like a reasonable solution.

nuttycom:
[force-pushed](https://github.com/Electric-Coin-Company/zcash-light-client-ffi/compare/c7ddd9da05096cd0e7361b5cf31495ef868e2a3c..1a54f0893d992b8c9cc320be3e4d9bac37031664) to address comments from review.

rust/src/lib.rs
```diff
@@ -355,36 +310,89 @@ pub unsafe extern "C" fn zcashlc_init_data_database(
...
+    account_name: *mut c_char,
+    key_source: *mut c_char,
     seed_fingerprint: [u8; 32],
-    account_index: u32,
+    hd_account_index: u32,
+}
+
+impl FfiAccount {
+    const NOT_FOUND: FfiAccount = FfiAccount {
+        uuid_bytes: [0u8; 16],
```
str4d, line 329: This works as a sentinel because this is not a valid UUID v4 encoding (possibly not a valid encoding for any UUID? I don't recall precisely, but at least for UUID v4 there are 6 bits that need to be set specifically), so it is distinguishable.

str4d (approved):
utACK 1a54f0893d992b8c9cc320be3e4d9bac37031664

PR #173, opened 2024-12-06
Add FFI methods for creating a transaction via PCZT proving & signing.

nuttycom:


daira (dismissed):
utACK

rust/src/lib.rs
```diff
@@ -2932,6 +2932,252 @@ pub unsafe extern "C" fn zcashlc_create_proposed_transactions(
...
+                account_id,
+                OvkPolicy::Sender,
+                &proposal,
+            )
+            .map_err(|e| anyhow!("Error creating PCZT from single-step proposal: {}", e))?;
+
+            Ok(FfiBoxedSlice::some(pczt.serialize()))
+        } else {
+            Err(anyhow!(
+                "Multi-step proposals are not yet supported for PCZT generation."
```
daira, line 3009: Should we actually be returning an array of PCZTs?

rust/src/lib.rs
```diff
@@ -2932,6 +2932,252 @@ pub unsafe extern "C" fn zcashlc_create_proposed_transactions(
...
+///
+/// - `db_data` must be non-null and valid for reads for `db_data_len` bytes, and it must
+///   have an alignment of `1`. Its contents must be a string representing a valid system
+///   path in the operating system's preferred representation.
+/// - The memory referenced by `db_data` must not be mutated for the duration of the
+///   function call.
+/// - The total size `db_data_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of `pointer::offset`.
+/// - `pczt_with_proofs_ptr` must be non-null and valid for reads for `pczt_with_proofs_len` bytes, and it
+///   must have an alignment of `1`. Its contents must be an encoded Proposal protobuf.
```
daira, line 3092: > Its contents must be an encoded Proposal protobuf.

I don't think that's a safety requirement?

rust/src/lib.rs
```diff
@@ -2932,6 +2932,252 @@ pub unsafe extern "C" fn zcashlc_create_proposed_transactions(
...
+/// - The memory referenced by `pczt_with_proofs_ptr` must not be mutated for the duration of the
+///   function call.
+/// - The total size `pczt_with_proofs_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of `pointer::offset`.
+/// - `pczt_with_sigs_ptr` must be non-null and valid for reads for `pczt_with_sigs_len` bytes, and it
+///   must have an alignment of `1`. Its contents must be an encoded Proposal protobuf.
```
daira, line 3098: Ditto.

rust/src/lib.rs
```diff
@@ -2932,6 +2932,252 @@ pub unsafe extern "C" fn zcashlc_create_proposed_transactions(
...
+///   function call.
+/// - The total size `db_data_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of `pointer::offset`.
+/// - `pczt_with_proofs_ptr` must be non-null and valid for reads for `pczt_with_proofs_len` bytes, and it
+///   must have an alignment of `1`. Its contents must be an encoded Proposal protobuf.
+/// - The memory referenced by `pczt_with_proofs_ptr` must not be mutated for the duration of the
+///   function call.
+/// - The total size `pczt_with_proofs_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of `pointer::offset`.
+/// - `pczt_with_sigs_ptr` must be non-null and valid for reads for `pczt_with_sigs_len` bytes, and it
```
daira, line 3097:
```suggestion
/// - `pczt_with_proofs_ptr` must be non-null and valid for reads for `pczt_with_proofs_len` bytes, and it
```

rust/src/lib.rs
```diff
@@ -2932,6 +2932,252 @@ pub unsafe extern "C" fn zcashlc_create_proposed_transactions(
...
+/// - The memory referenced by `pczt_with_sigs_ptr` must not be mutated for the duration of the
+///   function call.
+/// - The total size `pczt_with_sigs_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of `pointer::offset`.
+/// - `pczt_with_sigs_ptr` must be non-null and valid for reads for `pczt_with_sigs_len` bytes, and it
+///   must have an alignment of `1`. Its contents must be an encoded Proposal protobuf.
+/// - The memory referenced by `pczt_with_sigs_ptr` must not be mutated for the duration of the
+///   function call.
+/// - The total size `pczt_with_sigs_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of `pointer::offset`.
```
daira, line 3108: These lines seem to be a duplicate of the ones above.
```suggestion
```

rust/src/lib.rs
nuttycom: For a multi-step proposal, we will likely want a different API; there will be some cases where we can construct all of the PCZTs for the steps at once, but that's not the case for every multi-step proposal (for example, if the first step requires a multisig PCZT, I don't think we can necessarily construct the PCZT for the second step.)

rust/src/lib.rs
```diff
@@ -2932,6 +2932,252 @@ pub unsafe extern "C" fn zcashlc_create_proposed_transactions(
...
+/// - `spend_params` must be non-null and valid for reads for `spend_params_len` bytes,
+///   and it must have an alignment of `1`. Its contents must be the Sapling spend proving
+///   parameters.
```
daira, line 3111: > Its contents must be the Sapling spend proving parameters.

I don't think this is a safety requirement.

rust/src/lib.rs
```diff
@@ -2932,6 +2932,252 @@ pub unsafe extern "C" fn zcashlc_create_proposed_transactions(
...
+/// - The memory referenced by `spend_params` must not be mutated for the duration of the
+///   function call.
+/// - The total size `spend_params_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of `pointer::offset`.
+/// - `output_params` must be non-null and valid for reads for `output_params_len` bytes,
+///   and it must have an alignment of `1`. Its contents must be the Sapling output
+///   proving parameters.
```
daira, line 3118: > Its contents must be the Sapling output proving parameters.

I don't think this is a safety requirement.

rust/src/lib.rs
nuttycom: It's also wrong; I copied this documentation from elsewhere but this is a path. I was actually wondering whether this should be passed as a CStr instead.

daira (approved):
utACK with comments. The comments on safety documentation are blocking (if I'm correct).

rust/src/lib.rs
daira: We have other places where we pass a path to parameter files and it should match those, I think.

nuttycom: I'm not sure why those other paths aren't passed as C strings.

rust/src/lib.rs
nuttycom: Note that although it's not a safety requirement, we include what the contents are supposed to be in the safety documentation everywhere (see also the path to `db_data` etc.). Duplicating the parameter documentation seems excessive to me.

rust/src/lib.rs
daira: Actually, loading in the parameters from a file is quite expensive iirc. Don't we have a way to preload them?

daira (approved):
utACK

PR #174, opened 2024-12-09
Add methods for deriving addresses from UFVK and UIVK values to the `derivation` module.

nuttycom:


rust/src/derivation.rs
```diff
@@ -374,6 +376,130 @@ impl zcash_address::TryFromRawAddress for UnifiedAddressParser {
...
+    if !ptr.is_null() {
+        let ffi_address: Box<FfiAddress> = unsafe { Box::from_raw(ptr) };
+        if !(ffi_address.address.is_null()) {
+            unsafe { zcashlc_string_free(ffi_address.address) }
+        }
+        drop(ffi_address);
+    }
+}
+
+/// Derives a unified address address for the provided UFVK, along with the diversifier at which it
```
daira, line 417: Would we ever need a `zcashlc_derive_uivk_from_ufvk`? I ask because that's not expressible using these two functions.

rust/src/derivation.rs
```diff
@@ -374,6 +376,130 @@ impl zcash_address::TryFromRawAddress for UnifiedAddressParser {
...
+///
+/// # Safety
+///
+/// - `ufvk` must be non-null and must point to a null-terminated UTF-8 string.
+/// - `diversifier_index_bytes must either be null or be valid for reads for 11 bytes and have an
+///   alignment of `1`.
+/// - Call [`zcashlc_free_ffi_address`] to free the memory associated with the returned pointer
+///   when done using it.
+#[no_mangle]
+pub unsafe extern "C" fn zcashlc_derive_address_ufvk(
```
daira, line 430:
```suggestion
pub unsafe extern "C" fn zcashlc_derive_address_from_ufvk(
```

rust/src/derivation.rs
```diff
@@ -374,6 +376,130 @@ impl zcash_address::TryFromRawAddress for UnifiedAddressParser {
...
+///
+/// # Safety
+///
+/// - `uivk` must be non-null and must point to a null-terminated UTF-8 string.
+/// - `diversifier_index_bytes must either be null or be valid for reads for 11 bytes and have an
+///   alignment of `1`.
+/// - Call [`zcashlc_string_free`] to free the memory associated with the returned pointer
+///   when done using it.
+#[no_mangle]
+pub unsafe extern "C" fn zcashlc_derive_address_uivk(
```
daira, line 473:
```suggestion
pub unsafe extern "C" fn zcashlc_derive_address_from_uivk(
```

daira (approved):
utACK with non-blocking comments.

rust/src/derivation.rs
nuttycom: We should likely expose it, yes; the derivation tool should provide this sort of functionality.

nuttycom: Opened https://github.com/Electric-Coin-Company/zcash-light-client-ffi/issues/175

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- `zcashlc_derive_address_ufvk`
```
daira, line 10: These should have `from_` added to the name.

PR #178, opened 2024-12-11
Move off of the deprecated macos-12 builder.

nuttycom:


.github/workflows/build-binary-preview.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   build:
     name: Build
-    runs-on: macos-12
+    runs-on: macos-13.3
```
str4d, line 12: [This does not exist](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).
```suggestion
    runs-on: macos-13
```

str4d (changes requested):
Reviewed 2b19a7aa64120e9729cc72cfbe48b6a314878f63.

str4d (approved):

PR #190, opened 2025-02-13
CI: Improvements to binary preview workflow

str4d:


zizmor.yml
```diff
@@ -0,0 +1,4 @@
+rules:
+  cache-poisoning:
+    ignore:
+      - build-binary-preview.yml
```
daira, line 4: What is the potential cache poisoning attack that we're ignoring a warning about? Is it definitely a false positive?

PR #193, opened 2025-02-14
Add support for fetching and submitting transactions over Tor

str4d:


rust/src/lib.rs
```diff
@@ -3548,6 +3548,46 @@ pub unsafe extern "C" fn zcashlc_free_tor_runtime(ptr: *mut TorRuntime) {
     }
 }
 
+/// Returns a new isolated `TorRuntime` handle.
+///
+/// The two `TorRuntime`s will share internal state and configuration, but their streams
+/// will never share circuits with one another.
```
daira, line 3554: Missed opportunity for a "don't cross the streams!" joke ;-)

rust/src/lib.rs
```diff
@@ -3548,6 +3548,46 @@ pub unsafe extern "C" fn zcashlc_free_tor_runtime(ptr: *mut TorRuntime) {
...
+/// `TorRuntime` handle, but where you don't want their activities to be linkable to one
+/// another over the Tor network.
+///
+/// Calling this method is usually preferable to creating a completely separate
+/// `TorRuntime` instance, since it can share its internals with the existing `TorRuntime`.
+///
+/// # Safety
+///
+/// - `tor_runtime` must be non-null and point to a struct having the layout of
+///   [`TorRuntime`].
```
daira, line 3566: Is that sufficient? I don't think it would be ok to `memcpy` a `TorRuntime` and pass a pointer to the copy; I *think* it has to be a pointer returned by an API function with return type `*mut TorRuntime`.

daira (commented):
Blocking comments about safety documentation.

rust/src/lib.rs
str4d: The safety reqirements here are those necessary for the `unsafe { tor_runtime.as_mut() }` below to be safe. In particular, if non-null it must be [convertible to a reference](https://doc.rust-lang.org/stable/std/ptr/index.html#pointer-to-reference-conversion).

AFAICT it would be perfectly fine to `memcpy` a `TorRuntime` as long as you then discard the old one, which in Rust would be equivalent to a move. If you `memcpy` and then attempt to use both, that would be equivalent to `impl Copy for TorRuntime` which is disallowed because `Arc` does not implement `Copy` (because any copy cannot be purely bitwise due to also needing to update reference counters).

But that has no bearing on this method: as long as it can construct a valid `&mut TorRuntime` then it will work as long as the underlying `TorRuntime` has the correct layout (as per the safety requirements). The problem you envisage specifically arises if you:
- Obtain a `*mut TorRuntime`
- `memcpy` it in some way to obtain a second `*mut TorRuntime`.
- Call `zcashlc_free_tor_runtime` with both, which will cause a double-free (or more likely other problems because the `memcpy`ed one was not allocated with the Rust allocator).

So I think the only change I'd make here is independent of this PR: document on all `zcashlc_free_*` methods where their arguments must have come from.

rust/src/lib.rs
```diff
@@ -3548,6 +3548,46 @@ pub unsafe extern "C" fn zcashlc_free_tor_runtime(ptr: *mut TorRuntime) {
...
+/// - Call [`zcashlc_free_tor_runtime`] to free the memory associated with the returned
+///   pointer when done using it.
+#[no_mangle]
+pub unsafe extern "C" fn zcashlc_tor_isolated_client(
+    tor_runtime: *mut TorRuntime,
+) -> *mut TorRuntime {
+    // SAFETY: We ensure unwind safety by:
+    // - using `*mut TorRuntime` and respecting mutability rules on the Swift side, to
+    //   avoid observing the effects of a panic in another thread.
+    // - discarding the `TorRuntime` whenever we get an error that is due to a panic.
```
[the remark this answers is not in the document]
str4d: This safety documentation comes from the PR in which I added exchange rate fetching (#142). Unfortunately I can't find where I determined that this was the way to ensure unwind safety for `*mut T`. AFAICT the Swift caller is not discarding the `TorRuntime` [in existing usage](https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/blob/69dc9e9cb9df7e3592f7a8245a421698717c5232/Sources/ZcashLightClientKit/Tor/TorClient.swift#L40-L45) when getting an error that is due to a panic; I don't recall whether I meant here that `tor_runtime` is not used outside of `catch_panic`, or whether I did indeed intend for us to treat it that way in Swift and we're not doing so. The fact that I referenced Swift in the first point but not the second means I might have meant the former?

In any case, what we're doing here is the same as what we're doing everywhere else, so I've opened #194 for reviewing what we currently do and ensuring it is correct, or fixing it if not.

rust/src/lib.rs
daira: > AFAICT it would be perfectly fine to memcpy a TorRuntime as long as you then discard the old one, which in Rust would be equivalent to a move.

Agreed, it's only using the copy that is unsafe. But I don't agree that you can destroy the original and safely use the copy. That's not safe in general for opaque types. For example it would be UB to use `with_exposed_provenance[_mut]` to reconstruct a pointer into such a type, after exposing a pointer into the original. (Regardless of whether the current code does this, it *might* need to do it in future.)

rust/src/lib.rs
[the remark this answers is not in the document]
daira: > AFAICT the Swift caller is not discarding the TorRuntime in existing usage when getting an error that is due to a panic [...]

Indeed, and I think that means it is not panic-safe. The invariants of the `TorRuntime` or `LwdConn` could have been invalidated by whatever caused the panic.

rust/src/lib.rs
daira: Contrast with https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/196/files#r1966356159 where the "must point to a struct having the layout of" wording *is* sufficient.

rust/src/lib.rs
```diff
@@ -3548,6 +3548,46 @@ pub unsafe extern "C" fn zcashlc_free_tor_runtime(ptr: *mut TorRuntime) {
...
+    let tor_runtime = AssertUnwindSafe(tor_runtime);
+
+    let res = catch_panic(|| {
+        let tor_runtime =
+            unsafe { tor_runtime.as_mut() }.ok_or_else(|| anyhow!("A Tor runtime is required"))?;
+
+        let isolated_client = tor_runtime.isolated_client();
+
+        Ok(Box::into_raw(Box::new(isolated_client)))
+    });
```
daira, line 3587: https://github.com/Electric-Coin-Company/zcash-light-client-ffi/issues/194#issuecomment-2682144109

str4d:
[Rebased on `release/0.13.0`](https://github.com/Electric-Coin-Company/zcash-light-client-ffi/compare/1567df6e2ab17fa6568ca598ad836decaad773c1..d1781abfd891e1c8b795bdcce001ff3dc0564bc9) to fix merge conflicts and address review comments.

str4d:
Tested and working in the Swift SDK.

str4d:
Merging so we can get the Swift SDK side up, and the remaining review comments go beyond the bounds of this PR and might require altering other APIs (so we should handle them together in a separate PR that closes #194).

str4d:
(Actually, holding off merging until I debug and fix the release preview builds.)

str4d:
CI finally fixed.

PR #196, opened 2025-02-21
Add derivation methods for ZIP 325 private use metadata keys

str4d:


releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
str4d, line 1: I committed these header changes because I wanted to confirm my refactor only changed documentation, not ABIs. There are still pending binary changes that are being handled by the preview builds in the usual way.

rust/src/derivation.rs
```diff
@@ -579,6 +540,175 @@ pub unsafe extern "C" fn zcashlc_get_sapling_receiver_for_unified_address(
     unwrap_exc_or_null(res)
 }
 
+/// Constructs an ffi::AccountMetadataKey from its parts.
+///
+/// # Safety
+///
+/// - `sk` must be non-null and valid for reads for 32 bytes, and it must have an alignment of `1`.
+/// - `chain_code` must be non-null and valid for reads for 32 bytes, and it must have an alignment
+///   of `1`.
```
daira, line 549:
```suggestion
/// - `sk` must be non-null and valid for reads for 32 bytes, and it must have an alignment of `1`.
/// - The memory referenced by `sk` must not be mutated for the duration of the function call.
/// - `chain_code` must be non-null and valid for reads for 32 bytes, and it must have an alignment
///   of `1`.
/// - The memory referenced by `chain_code` must not be mutated for the duration of the function
///   call.
```

rust/src/derivation.rs
```diff
@@ -579,6 +540,175 @@ pub unsafe extern "C" fn zcashlc_get_sapling_receiver_for_unified_address(
...
+) -> *mut ffi::AccountMetadataKey {
+    let res = catch_panic(|| {
+        let network = parse_network(network_id)?;
+        let seed = unsafe { slice::from_raw_parts(seed, seed_len) };
+        let account = zip32_account_index(account)?;
+
+        let key = zip32::registered::SecretKey::from_subpath(
+            b"MetadataKeys",
+            seed,
+            // TODO: Change this to whatever ZIP number is assigned to the metadata key ZIP draft.
```
daira, line 597:
```suggestion
```
This was already assigned in a meeting that had a sufficient ZIP Editor quorum.

rust/src/derivation.rs
```diff
@@ -579,6 +540,175 @@ pub unsafe extern "C" fn zcashlc_get_sapling_receiver_for_unified_address(
...
+/// - For encryption-like private usage, the first key in the array should always be
+///   used, and all other keys ignored.
+/// - For decryption-like private usage, each key in the array should be tried in turn
+///   until metadata can be recovered, and then the metadata should be re-encrypted
+///   under the first key.
+///
+/// # Safety
+///
+/// - `account_metadata_key` must be non-null and must point to a struct having the layout of
+///   [`ffi::AccountMetadataKey`].
```
daira, line 633: Note that the objection in https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/193#discussion_r1957093909 does *not* apply here because `ffi::AccountMetadataKey` is a non-opaque, "plain old data" type: we define the layout, we know that it has no internal pointers, and we have no need to discard / reconstruct the provenance of the pointer because it is not passed to any API that retains it.

daira (approved):
I did not check for changes in the moved code. Otherwise utACK

rust/src/derivation.rs
daira: There might be other missing cases of "The memory referenced by `...` must not be mutated for the duration of the function call." I didn't check.

str4d: Technically we don't need it to be unmutated for the duration of the function call, as we read the pointer as a slice and then immediately convert into a stack-allocated array (thus no longer depending on the original pointer's memory). Compare this to `seed` where we read the pointer as a slice, and then pass that slice into other functions.

That being said, conveying a requirement that the memory not be modified for "the first few operations within the function call" is hard, so I'm fine with having a much stricter requirement than we actually need.

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-light-client-ffi/compare/74b24b36e4cd6c5790624763c769572ca2c48c8e..2bc682ac7805037016383103cdd6de692ee61d69) to address review comments.

rust/src/derivation.rs
```diff
@@ -579,6 +540,177 @@ pub unsafe extern "C" fn zcashlc_get_sapling_receiver_for_unified_address(
...
+/// - If `ufvk` is non-null, it must point to a null-terminated UTF-8 string.
+/// - `private_use_subject` must be non-null and valid for reads for `private_use_subject_len`
+///   bytes.
+/// - The memory referenced by `private_use_subject` must not be mutated for the duration
+///   of the function call.
+/// - The total size `private_use_subject_len` must be no larger than `isize::MAX`. See
+///   the safety documentation of `pointer::offset`.
+/// - Call `zcashlc_free_boxed_slice` to free the memory associated with the returned
```
str4d, line 643:
```suggestion
/// - Call `zcashlc_free_txids` to free the memory associated with the returned
```

Also I still need to experiment to see if I can create aliases in Rust that are exposed as separate types in the C FFI (so I can rename `ffi::TxIds` for this new more general usage).

rust/src/derivation.rs
```diff
@@ -432,7 +393,7 @@ pub unsafe extern "C" fn zcashlc_derive_address_from_ufvk(
     network_id: u32,
     ufvk: *const c_char,
```
daira, line 394: Add "The memory referenced by `ufvk` must not be mutated for the duration of the function call."

daira (approved):
utACK with comments.

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-light-client-ffi/compare/2bc682ac7805037016383103cdd6de692ee61d69..9b3fcb1300800e280f82e1ba2d30784a607dfad2) to address remaining comments on this PR's contents, and fix some bugs.

str4d:
And [one last force-push](https://github.com/Electric-Coin-Company/zcash-light-client-ffi/compare/9b3fcb1300800e280f82e1ba2d30784a607dfad2..50bc34b840c1ab3a50fca1b91cd74983f9e1d429) to fix the changelog.

I've tested this with https://github.com/zcash/zcash-test-vectors/pull/106 in the Swift SDK; this is ready to merge.

PR #197, opened 2025-02-27
Fix binary preview

str4d:
This is necessary to discover any existing binary previews for the current branch, so we can merge to it instead of overwriting it.


str4d:
Not quite sure if it's working; the third commit does check out the existing branch, but presumably because there were no code changes in the PR it showed as nothing changed after the build. I'll merge this and then see what happens after I merge the Tor PR.

daira (commented):
Post-hoc ACK for #197 and #198.

PR #199, opened 2025-02-28
Expose mined height from `zcashlc_tor_lwd_conn_fetch_transaction`

str4d:
This was intended to be in Electric-Coin-Company/zcash-light-client-ffi#193 but I forgot to push it to the PR before merging the latter.


str4d:
Self-ACK, this code was tested locally (that was part of the problem, I wrote the Swift side assuming this was in, and then the binary preview never included it).

daira (commented):
Post-hoc ACK

PR #200, opened 2025-02-28
CI: Enable general Rust caching

str4d:


str4d:
This failed to enable sccache writing, but it's still probably useful.

PR #202, opened 2025-03-01
Expose more lightwalletd gRPCs over Tor

str4d:


rust/src/ffi.rs
```diff
@@ -878,3 +878,97 @@ pub unsafe extern "C" fn zcashlc_free_account_metadata_key(ptr: *mut AccountMeta
         drop(key);
     }
 }
+
+/// Information about a lightwalletd instance and the state of the blockchain.
+#[repr(C)]
+pub struct LightdInfo {
```
nuttycom, line 884: Question: why are we making a C type for the purpose of pushing something that's already encoded as a protobuf over the FFI? Let's just pass the protobuf.

nuttycom (approved):
utACK with nonblocking (but maybe worth changing?) comment.

rust/src/ffi.rs
str4d: That's a very good point, I completely forgot about that point and instead explicitly parsed the protobuf here and then reconstruct it on the Swift side. I will fix this 😅

rust/src/tor.rs
```diff
@@ -81,6 +82,34 @@ pub struct LwdConn {
 }
 
 impl LwdConn {
+    /// Returns information about this lightwalletd instance and the blockchain.
+    pub(crate) fn get_lightd_info(&mut self) -> anyhow::Result<service::LightdInfo> {
+        Ok(self
+            .runtime
+            .clone()
+            .block_on(async { self.conn.get_lightd_info(service::Empty {}).await })?
```
nuttycom, line 90: Rust question: it's surprising to me that the `async` and `await` are required here. My mental model of async functions is that they're just syntactic sugar over functions that return their results in a `Future`; if that were the case though, with `.block_on` taking an argument of type `Future<A>` and `.get_lightd_info` returning a Future, why the need to wrap the call?

nuttycom (approved):
utACK 39eecf874c3600b633068072d0665820ae1f2381

rust/src/tor.rs
str4d: You're correct that this could be simplified. I'm just in the habit of the `async {}` wrapping because usually I'm invoking more than one async method.

nuttycom (approved):
re-utACK 0482903e65f8644a5b56dc4e3b52662189f61c75

nuttycom:
Looks like the preview build is due to an ccache issue? Copying the relevant error here in case the log disappears:
```sccache: error: Server startup failed: cache storage failed to read: Unexpected (permanent) at read => {"$id":"1","innerException":null,"message":"This legacy service is shutting down, effective April 15, 2025. Migrate to the new service ASAP. For more information: https://gh.io/gha-cache-sunset","typeName":"Microsoft.Azure.DevOps.ArtifactCache.WebApi.ArtifactCacheServiceDecommissionedException, Microsoft.Azure.DevOps.ArtifactCache.WebApi","typeKey":"ArtifactCacheServiceDecommissionedException","errorCode":0,"eventId":3000}```

PR #203, opened 2025-03-04
Update the README to refer to version 0.13.0

nuttycom:


str4d:
You also forgot to change the version in Cargo.toml during the release process. Not the end of the world as we don't currently use that version value anywhere, but we should generally bump it for consistency.

str4d (approved):

PR #205, opened 2025-03-05
Rust updates and 2024 edition

str4d:


rust/src/lib.rs
```diff
@@ -779,7 +790,9 @@ pub unsafe extern "C" fn zcashlc_list_transparent_receivers(
         let db_data = unsafe { wallet_db(db_data, db_data_len, network)? };
         let account_uuid = account_uuid_from_bytes(account_uuid_bytes)?;
 
-        match db_data.get_transparent_receivers(account_uuid) {
+        // This API is documented as returning the transparent receivers for allocated
+        // diversified UAs, which means change taddrs are excluded.
+        match db_data.get_transparent_receivers(account_uuid, false) {
```
str4d, line 795: I think the intention is to eventually pass both external and change addresses over the same FFI call, or maybe expose this boolean across it. I'm not doing that in this PR as it's focused on the Rust updates and MSRV bump; I want this to be separately testable against the Swift SDK before we start making the transparent-related and address-rotation-related FFI changes (plus this is a big PR already).

nuttycom: This is the correct change for the current PR.

rust/src/lib.rs
```diff
@@ -888,7 +901,7 @@ pub unsafe extern "C" fn zcashlc_get_verified_transparent_balance_for_account(
             })
             .and_then(|target| {
                 db_data
-                    .get_transparent_receivers(account_uuid)
+                    .get_transparent_receivers(account_uuid, true)
```
nuttycom, line 904: This could also be set to `false` but it should be harmless in the context of Zashi; we will never have queried for funds sent to these taddrs, so no funds sent to those taddrs should show up in the balance.

rust/src/lib.rs
```diff
@@ -2137,12 +2151,12 @@ pub unsafe extern "C" fn zcashlc_propose_shielding(
...
             |addr| Ok(account_receivers.get(&addr).map(|value| (addr, *value)))
-        )?.filter(|(_, value)| *value >= shielding_threshold) {
+        )?.filter(|(_, value)| *value >= shielding_threshold) { Some((addr, _)) => {
             [addr]
-        } else {
+        } _ => {
             // There are no transparent funds to shield; don't create a proposal.
             return Ok(ffi::BoxedSlice::none());
-        };
+        }};
```
nuttycom, line 2159: Was this a clippy suggestion? It seems less clear to me.

nuttycom (approved):
utACK 93e1fca6431f2db459321578803a61946e9ae216

rust/src/lib.rs
str4d: Rust 2024 edition introduces rustfmt editions, to enable changing or fixing formatting issues. This is one such formatting change in the edition.

PR #206, opened 2025-03-07
Allow shielding from multiple transparent addresses.

nuttycom:


rust/src/lib.rs
```diff
@@ -792,7 +792,7 @@ pub unsafe extern "C" fn zcashlc_list_transparent_receivers(
 
         // This API is documented as returning the transparent receivers for allocated
         // diversified UAs, which means change taddrs are excluded.
-        match db_data.get_transparent_receivers(account_uuid, false) {
+        match db_data.get_transparent_receivers(account_uuid, true) {
```
str4d, line 795: If you're going to hard-code always exposing change addresses (rather than exposing the boolean as an FFI parameter), then the API documentation needs updating and every callsite on the Swift SDK side needs reviewing to ensure they all behave correctly with these new semantics.

In any case, the comment above this line (documenting why this was `false` before) needs updating.

rust/src/lib.rs
```diff
@@ -2142,24 +2142,25 @@ pub unsafe extern "C" fn zcashlc_propose_shielding(
...
-            } else {
-                Ok(account_receivers.iter().next().map(|(a, v)| (*a, *v)))
-            },
-            |addr| Ok(account_receivers.get(&addr).map(|value| (addr, *value)))
-        )?.filter(|(_, value)| *value >= shielding_threshold) { Some((addr, _)) => {
-            [addr]
-        } _ => {
-            // There are no transparent funds to shield; don't create a proposal.
+        // If a specific receiver is specified, select only value for that receiver; otherwise,
+        // select value for all receivers.
```
str4d, line 2146: Document why we are selecting all receivers where previously we only selected one (because we're relying on the assumption that any transparent receivers with funds in them were either already linked together by transparent-only wallets, or are newly linked together by mobile SDK usage by choice for short-term usability, and that the transparent gap limits being used are low enough that we can separate these now-definitely-linked addresses from future unlinked addresses by raising the gap limit if/when we deploy a mechanism that enables unlinkable transparent detection).

str4d (changes requested):
Reviewed 45c14d53fb1b8ed4549e51a62cdf29e856e51b23. Blocking on documentation (or possibly an API change in one case).

daira (commented):
No additional comments to @str4d's.

nuttycom:
Closing in favor of #207 (to enable automatic preview build of binary artifacts.)

PR #207, opened 2025-03-07
Allow shielding from multiple transparent addresses.

nuttycom:
Replaces #206 (branch name has changed so that an automatic preview build can be created.)

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -792,7 +792,7 @@ pub unsafe extern "C" fn zcashlc_list_transparent_receivers(
 
         // This API is documented as returning the transparent receivers for allocated
         // diversified UAs, which means change taddrs are excluded.
-        match db_data.get_transparent_receivers(account_uuid, false) {
+        match db_data.get_transparent_receivers(account_uuid, true) {
             Ok(receivers) => {
                 let keys = receivers
                     .keys()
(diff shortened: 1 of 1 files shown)


rust/src/lib.rs
```diff
@@ -2142,24 +2162,27 @@ pub unsafe extern "C" fn zcashlc_propose_shielding(
...
-            },
-            |addr| Ok(account_receivers.get(&addr).map(|value| (addr, *value)))
-        )?.filter(|(_, value)| *value >= shielding_threshold) { Some((addr, _)) => {
-            [addr]
-        } _ => {
-            // There are no transparent funds to shield; don't create a proposal.
+        // If a specific receiver is specified, select only value for that receiver; otherwise,
+        // select value for all receivers. See the warnings associated with the documentation
+        // of the `transparent_receiver` argument in the method documentation for privacy
+        // considerations.
```
daira, line 2168: As a general rule I prefer that if a semantic change could potentially reduce security for existing callers (in any situation), it is deliberately API/ABI-breaking. In this particular case the Swift SDK is basically the only user of the API so I'm not going to insist on it.

rust/src/lib.rs
```diff
@@ -2054,6 +2053,27 @@ pub unsafe extern "C" fn zcashlc_string_free(s: *mut c_char) {
...
 /// with `zcashlc_create_proposed_transaction`.
 ///
+/// # Parameters
+///
+/// - db_data: A string represented as a sequence of UTF-8 bytes.
+/// - db_data_len: The length of `db_data`, in bytes.
+/// - account_uuid_bytes: a 16-byte array representing the UUID for an account
+/// - memo: `null` to represent "no memo", or a pointer to an array containing exactly 512 bytes.
+/// - shielding_threshold: the minimum value to be shielded; if insufficient value is available,
+///   this will fail with an error.
```
daira, line 2063:
```suggestion
/// - shielding_threshold: the minimum value to be shielded for each receiver.
```

rust/src/lib.rs
```diff
@@ -2054,6 +2053,27 @@ pub unsafe extern "C" fn zcashlc_string_free(s: *mut c_char) {
 /// transaction that can then be authorized and made ready for submission to the network
 /// with `zcashlc_create_proposed_transaction`.
```
daira, line 2054:
```suggestion
/// with `zcashlc_create_proposed_transaction`. If there are no receivers (as
/// selected by `transparent_receiver`) for which at least `shielding_threshold`
/// of value is available to shield, fail with an error.
```

daira (approved):
utACK with documentation suggestions.

PR #218, opened 2025-05-12
Add the ability to specify custom unified addresses.

nuttycom:


rust/src/lib.rs
```diff
@@ -739,15 +783,22 @@ pub unsafe extern "C" fn zcashlc_get_next_available_address(
     db_data_len: usize,
     account_uuid_bytes: *const u8,
     network_id: u32,
+    receiver_flags: u8,
```
str4d, line 786: I'm a little uncomfortable adding a `u8` bitflag with half of its values already assigned. I can imagine a minimum of two more kinds of receivers we would want in future (long-term storage, and PQ-safe), potentially more if we add receiver types with e.g. detection keys. I would suggest making this at least a `u16` from the get-go.

rust/src/lib.rs
```diff
@@ -739,15 +783,22 @@ pub unsafe extern "C" fn zcashlc_get_next_available_address(
...
+            .ok_or_else(|| anyhow!("Invalid unified address receiver flags {}", receiver_flags))?;
+        let address_request = receiver_flags.to_address_request().map_err(|_| {
+            anyhow!(
+                "Could not generate a valid unified address for flags {}",
+                receiver_flags.bits()
+            )
+        })?;
 
-        match db_data
-            .get_next_available_address(account_uuid, UnifiedAddressRequest::AllAvailableKeys)
```
str4d, line 749: I confirmed that the previous value can be expressed with the new API, as it happens to only use `Omit` and `Require`, and the choice of these is fixed in our compiled FFI binary.

rust/src/lib.rs
```diff
@@ -716,8 +717,51 @@ pub unsafe extern "C" fn zcashlc_get_current_address(
...
+    }
+}
+
+impl ReceiverFlags {
+    fn to_address_request(&self) -> Result<UnifiedAddressRequest, ()> {
+        UnifiedAddressRequest::custom(
+            if self.contains(ReceiverFlags::ORCHARD) {
+                ReceiverRequirement::Require
+            } else {
+                ReceiverRequirement::Omit
```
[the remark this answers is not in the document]
nuttycom: You don't.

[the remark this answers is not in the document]
nuttycom: `Allow` is not really that useful at the SDK level; it's not something you'd ever expose to a user, and an SDK caller should always either require or omit.

rust/src/lib.rs
nuttycom: In that case I'll just go to u32, since it will come across the FFI as i32 anyway.

nuttycom: Oh wait, I'm thinking of Android. But in any case, I'll use u32 for parity with Android.

rust/src/lib.rs
```diff
@@ -716,8 +717,56 @@ pub unsafe extern "C" fn zcashlc_get_current_address(
...
+
 /// Returns a newly-generated unified payment address for the specified account, with the next
-/// available diversifier.
+/// available diversifier and the specified set of receivers.
+///
+/// The set of receivers to include in the generated address is specified by a byte which may have
+/// any of the following bits set:
+/// * P2PKH = 0b00000001
+/// * SAPLING = 0b00000100
+/// * ORCHARD = 0b00001000
```
str4d, line 765: Nit: now that the bit length is not `u8`, these might be clearer either with the full set of prefixing zeroes (so there is no possible ambiguity in where the extra zero bits go), or as `(1 << 0)` etc. so we're just stating the relevant bit position.

str4d (approved):
utACK 615372cd2dc8ac12a80f1824164a4db17657049b

PR #222, opened 2025-05-22
Bring in latest crate revision with reliability improvements

str4d:
Reopened #217 after that PR was backed out of 0.16.0.


nuttycom (approved):
utACK

PR #236, opened 2025-09-01
feature: librustzcash update confirmations policy part 1

schell:
This updates the API to use `librustzcash` 776023 which includes `ConfirmationsPolicy`. 

* bumps `librustzcash`
* reworks underlying FFI functions in terms of `ConfirmationsPolicy`
* updates deprecated `TxOut` field use

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -2810,29 +2810,30 @@ pub unsafe extern "C" fn zcashlc_transaction_data_requests(
                     TransactionDataRequest::Enhancement(txid) => {
                         ffi::TransactionDataRequest::Enhancement(txid.into())
                     }
-                    TransactionDataRequest::TransactionsInvolvingAddress {
-                        address,
-                        block_range_start,
-                        block_range_end,
-                        request_at,
-                        tx_status_filter,
-                        output_status_filter,
-                    } => ffi::TransactionDataRequest::TransactionsInvolvingAddress {
-                        address: CString::new(address.encode(&network)).unwrap().into_raw(),
-                        block_range_start: block_range_start.into(),
-                        block_range_end: block_range_end.map_or(-1, |h| u32::from(h).into()),
-                        request_at: request_at.map_or(-1, |t| {
-                            t.duration_since(UNIX_EPOCH)
-                                .expect("SystemTime should never be before the epoch")
-                                .as_secs()
-                                .try_into()
-                                .expect("we have time before a SystemTime overflows i64")
-                        }),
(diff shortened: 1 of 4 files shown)


rust/src/lib.rs
```diff
@@ -1072,26 +1060,25 @@ pub unsafe extern "C" fn zcashlc_get_total_transparent_balance_for_account(
...
+            .map_err(|e| {
+                anyhow!(
+                    "Error while fetching transparent balances for {:?}: {}",
+                    account_uuid,
+                    e,
+                )
+            })?;
+        let amount = balances
             .values()
+            // TODO(schell): check to see if `total` is the correct thing to use here
```
pacu, line 1080: I think this is ok.

cc'ing @nuttycom just to double-check

rust/src/lib.rs
```diff
@@ -2003,14 +2001,91 @@ pub unsafe extern "C" fn zcashlc_propose_transfer(
...
+
+        let proposal = propose_send_max_transfer::<_, _, _, Infallible>(
+            &mut db_data,
+            &network,
+            account_uuid,
+            &[ShieldedProtocol::Sapling, ShieldedProtocol::Orchard],
+            &StandardFeeRule::Zip317,
+            to,
+            memo,
+            MaxSpendMode::MaxSpendable,
```
pacu, line 2087: I think for Zashi this would be MaxSpendMode::Everything.

@nuttycom could you confirm this?

pacu (approved):
utACK: modulo questions for @nuttycom

rust/src/lib.rs
```diff
@@ -1072,26 +1059,25 @@ pub unsafe extern "C" fn zcashlc_get_total_transparent_balance_for_account(
...
+                anyhow!(
+                    "Error while fetching transparent balances for {:?}: {}",
+                    account_uuid,
+                    e,
+                )
+            })?;
+        let amount = balances
             .values()
+            // TODO(schell): check to see if `total` is the correct thing to use here
+            .map(|balance| balance.total())
```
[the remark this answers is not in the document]
schell: There's nothing in `Balance` regarding shielded vs transparent funds, so I'll assume that the function to use here instead of `.total()` is `.spendable_value()`.

nuttycom (approved):
utACK f15806b84c5e5d5cad1f666726c67ef81fdf4a9b. We believe that `zcashlc_get_total_transparent_balance_for_account` is unused in the Swift SDK and is not exposed via the public API, so we'll just delete it (in another PR, since this one is just updating to the latest `librustzcash`).

PR #237, opened 2025-09-02
Document the general development and release process

str4d:
New PR because #234 was merged a bit too eagerly.


nuttycom (approved):
doc-only ACK

PR #238, opened 2025-09-07
feature: librustzcash update confirmations policy part 2

schell:
This PR stacks on #236 and changes the FFI API to include `ConfirmationsPolicy`.

### DIFF
--- a/releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
+++ b/releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
@@ -173,13 +173,19 @@ typedef struct FFIEncodedKeys {
  */
 typedef struct ConfirmationsPolicy {
   /**
-   * NonZero, zero for default
+   * The number of confirmations required before trusted notes may be spent. NonZero, set this
+   * and `untrusted` to zero to accept the default value for each.
    */
   uint32_t trusted;
   /**
-   * NonZero, zero for default, zero must match `trusted`
+   * The number of confirmations required before untrusted notes may be spent. NonZero, set this
+   * and `trusted` both to zero to accept the default value for each.
    */
   uint32_t untrusted;
+  /**
+   * A flag that enables selection of zero-conf transparent UTXOs for spends in shielding
+   * transactions.
+   */
   bool allow_zero_conf_shielding;
 } ConfirmationsPolicy;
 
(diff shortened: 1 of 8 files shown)


nuttycom (changes requested):
One more change that should go into this PR is that you should run `make xcframework` from the root directory and commit the changes to the header files so that they're easy to review. If you don't have an OSX machine, I can do so.

schell:
> One more change that should go into this PR is that you should run `make xcframework` from the root directory and commit the changes to the header files so that they're easy to review. If you don't have an OSX machine, I can do so.

Thanks @nuttycom , will do.

nuttycom:
@schell I [force-pushed](https://github.com/Electric-Coin-Company/zcash-light-client-ffi/compare/d738b390cb00015f9d7e2b20face3b0477f67f56..8c011d9eb9c805df206a97a9b87ee9a3b1125f1e) this to remove the updates to the build artifacts; all I had wanted regenerated was the header files. The build artifacts themselves will only be regenerated when we actually cut the release, to prevent as much bloating of the git repository as is possible.

releases/XCFramework/libzcashlc.xcframework/ios-arm64/libzcashlc.framework/Headers/zcashlc.h
```diff
@@ -157,6 +157,32 @@ typedef struct FFIEncodedKeys {
...
+typedef struct ConfirmationsPolicy {
+  /**
+   * NonZero, zero for default
+   */
+  uint32_t trusted;
+  /**
+   * NonZero, zero for default, zero must match `trusted`
+   */
+  uint32_t untrusted;
+  bool allow_zero_conf_shielding;
```
nuttycom, line 183: Needs documentation in the sources that generate this data structure.

rust/src/ffi.rs
```diff
@@ -1081,3 +1081,78 @@ pub unsafe extern "C" fn zcashlc_free_http_response_bytes(ptr: *mut HttpResponse
...
+/// `0` will be treated as a request for a default value.
+///
+/// [`ZIP 315`]: https://zips.z.cash/zip-0315
+#[repr(C)]
+pub struct ConfirmationsPolicy {
+    /// NonZero, zero for default
+    pub(crate) trusted: u32,
+    /// NonZero, zero for default, zero must match `trusted`
+    pub(crate) untrusted: u32,
+    pub(crate) allow_zero_conf_shielding: bool,
```
nuttycom, line 1103: Document this.

PR #242, opened 2025-09-20
Return the txid of the decrypted transaction from zcashlc_decrypt_and_store_transaction

nuttycom:


rust/src/lib.rs
```diff
@@ -1846,6 +1846,8 @@ pub unsafe extern "C" fn zcashlc_latest_cached_block_height(
 /// - The memory referenced by `tx` must not be mutated for the duration of the function call.
 /// - The total size `tx_len` must be no larger than `isize::MAX`. See the safety
 ///   documentation of pointer::offset.
+/// - `txid_ret` must be non-null and must point to an allocated 32-byte region of memory. On
+///   successful execution this will contain the txid of the decrypted transaction.
```
daira, line 1850:
```suggestion
/// - `txid_ret` must be non-null and valid for writes of 32 bytes with an alignment of 1.
///   On successful execution (return value of 1) this will contain the txid of the decrypted
///   transaction.
```
At line 1836 add:
```rust
/// Returns 1 on success or -1 on failure.
///
```

daira (approved):
utACK with documentation suggestions.

rust/src/lib.rs
```diff
@@ -1886,7 +1889,10 @@ pub unsafe extern "C" fn zcashlc_decrypt_and_store_transaction(
         };
 
         match decrypt_and_store_transaction(&network, &mut db_data, &tx, mined_height) {
-            Ok(()) => Ok(1),
+            Ok(()) => {
+                unsafe { txid_ret.copy_from(tx.txid().as_ref().as_ptr(), 32) };
```
str4d, line 1893: This isn't how I normally do this kind of copy, but it looks fine.

str4d (commented):
Post-hoc utACK 59b1284380ad83dfe5d024bfdfae9a7f6e7c4bd5

